PR #1, opened 2023-07-13
BEM / @extend refactor basis

therealcoder1337:
As we talked about, this enables us to use our own CSS classes, using BEM and `@extend`. There's also an example refactor of QBtn.
Since this uses a Vite Plugin I wrote, it should work automatically on dev server startup and would also auto update the file whenever we bump the `beercss` dependency version.


therealcoder1337:
Not yet working as I hoped, it seems to include all the styles.

PR #5, opened 2023-07-26
QChip: BEM refactor

KilDesu:
- Refactored QChip so it follows BEM principles
- Added a btn-image mixin to be able to reuse the styles
- Added possibility to add image to QBtn with `icon="img:/cotcktail.jpg"`

### DIFF
--- a/src/lib/components/button/QBtn.svelte
+++ b/src/lib/components/button/QBtn.svelte
@@ -48,7 +48,11 @@
 >
   {#if icon && !loading}
     {#if icon.startsWith("img:")}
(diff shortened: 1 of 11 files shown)


src/lib/components/button/QBtn.svelte
```diff
@@ -47,7 +47,11 @@
   on:click
 >
   {#if icon && !loading}
-    <QIcon name={icon} class="q-btn__icon" />
+    {#if icon.startsWith("img:")}
+      <img src={icon.replace("img:", "")} class="q-btn__img q-btn__img--responsive" alt="" />
```
therealcoder1337, line 51: alt attribute missing a value.

therealcoder1337 (approved):
Good job!

PR #6, opened 2023-07-27
feat: QInput, QSelect: refactored to BEM. added mixins.

therealcoder1337:
Please note how I put some mixins as separate files in the newly created `mixins` directory. I think the `mixins.scss` would've gotten too big otherwise.


KilDesu (approved):
Looks good to me, good job!

PR #8, opened 2023-07-29
QTable BEM refactor & other changes

therealcoder1337:
- QTable now uses BEM
- Added new beer background image for light mode docs
- Fixed a few minor bugs


KilDesu (approved):
Good job, as always!

PR #11, opened 2023-07-31
feat: added content to home page and logo.

therealcoder1337:
WIP still, but better than an empty page. :)


KilDesu (approved):
Pretty cool, thanks!

PR #13, opened 2023-08-01
Layout and disable fixes + new action feature

KilDesu:
- Fixed a bug where the left drawer in main +layout.svelte didn't open in certain cases
- Added a fade transition when the layout drawer changes (e.g. whe you go from /components to /utils using left railbar)
- Previous point needed a change to layout styles
- Added a new action that allows to listen for click, enter key and space key to be used in several components like QBtn, QChip for example. It takes into account the disabled state of the component
- Refactored QBtn and QChip to use this action

### DIFF
--- a/src/routes/+layout.svelte
+++ b/src/routes/+layout.svelte
@@ -118,8 +118,6 @@
   ];
   let contentEl: HTMLDivElement;
 
-  $: console.log($Quaff.router.url.pathname);
-
   let selectedRailbarItem: "components" | "utils" | null;
   $: selectedRailbarItem = isRouteActive($Quaff.router, "/components")
     ? "components"
--- a/src/routes/+layout.svelte
+++ b/src/routes/+layout.svelte
@@ -18,6 +18,7 @@
     QDrawer,
   } from "$lib";
   import { isRouteActive } from "$lib/composables/use-router-link";
+  import { fade } from "svelte/transition";
 
   const components = [
     {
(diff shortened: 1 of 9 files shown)


src/routes/+layout.svelte
```diff
@@ -115,12 +116,12 @@
...
     },
   ];
-  let contentEl: HTMLDivElement | null = null;
+  let contentEl: HTMLDivElement;
 
-  let selectedRailbarItem: "components" | "utils" | null = isRouteActive(
-    $Quaff.router,
-    "/components"
-  )
+  $: console.log($Quaff.router.url.pathname);
```
therealcoder1337, line 121: Should be removed I guess

src/routes/dev/+page.svelte
```diff
@@ -177,12 +177,12 @@
     </div>
   </QCard>
   <QCard title="Buttons">
-    <QBtn class="q-ma-sm" icon="favorite" label="Using Label" />
+    <QBtn class="q-ma-sm" icon="favorite" label="Using Label" on:activated={console.log} />
```
therealcoder1337, line 180: This is intended, since this is on the dev page, right?

src/routes/dev/+page.svelte
```diff
@@ -177,12 +177,12 @@
...
     <QBtn class="q-ma-sm">
       <span>Using</span><span style="color: blue">slot</span>
     </QBtn>
     <QBtn class="q-ma-sm" label="loading" loading />
-    <QBtn class="q-ma-sm" icon="add" label="Disabled" disable />
+    <QBtn class="q-ma-sm" icon="add" label="Disabled" disable on:activated={console.log} />
```
therealcoder1337, line 185: Same here.

therealcoder1337 (approved):
Great! But please check my comments.

src/routes/dev/+page.svelte
KilDesu: Yes, it's just to show the on:activated event

PR #19, opened 2023-08-03
Fixes

therealcoder1337:
Regarding the `version`, it might be a bit over engineered to have a custom Vite plugin, but I couldn't get the `package.json` import to work (directory structure gets changed on build) and the previous solution didn't work either.
PrismJS caused several issues, especially the auto replacement of html contents which couldn't be disabled, causing buggy behavior (I tried `Prism.manual = true`, to no avail), that's why it was replaced by `svelte-highlight`.

### DIFF
--- a/src/lib/stores/Quaff.ts
+++ b/src/lib/stores/Quaff.ts
@@ -1,9 +1,10 @@
 import { writable, derived } from "svelte/store";
 import { page } from "$app/stores";
+import version from "$lib/helpers/version";
 
 function quaff() {
   const { subscribe, set, update } = writable({
-    version: __QUAFF_VERSION__,
+    version,
     dark: false,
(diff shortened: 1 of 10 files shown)


src/lib/stores/Quaff.ts
```diff
@@ -1,9 +1,10 @@
 import { writable, derived } from "svelte/store";
 import { page } from "$app/stores";
+import version from "$lib/helpers/version";
 
 function quaff() {
   const { subscribe, set, update } = writable({
-    version: __QUAFF_VERSION__,
```
KilDesu, line 6: The Quaff version should still be available on the $Quaff object, do you already have an idea how to get it?

KilDesu (approved):
Good job though I haven't been able to try those changes yet

PR #23, opened 2023-08-08
Refactors 08 aug

therealcoder1337:
Just some minor things. Spent most of the time debugging why `font-variation-settings` aren't considered for QCheckbox with the new font package, but didn't find the reason yet.


KilDesu (approved):
Good work!

PR #26, opened 2023-08-10
More refactors 09 aug

therealcoder1337:


src/lib/composables/use-align.ts
```diff
@@ -23,15 +23,25 @@ export const UseAlignPropsDefaults: UseAlignProps = {
...
-  const alignClass = align
+  const alignments = align
     .split(" ")
-    .map((a) => `${a}-align`)
-    .join(" ");
+    .map((entry) => {
+      const val = alignMap[entry as keyof typeof alignMap];
+      return val ? `justify-${val}` : false;
+    })
+    .filter((entry) => typeof entry === "string");
```
KilDesu, line 44: `.filter(Boolean)` ?

KilDesu (approved):
Just one comment. Good job

src/lib/composables/use-align.ts
therealcoder1337: It's an option, but does it really improve readability in regards to what the code does?

PR #27, opened 2023-08-10
refactor(size): created a better useSize function

KilDesu:
- Created a `useSize` function to handle all different sizes that could be received as input from users
- Wrote some types related to sizes
- Refactored some components to use the `useSize` function
- Added some default values for some components' props to be used in docs


therealcoder1337 (approved):
Thanks!

PR #28, opened 2023-08-11
Qlayout refactor

KilDesu:
Refactored QLayout. It's been a real hassle so expect some bugs. No bugs were found yet, even without BeerCSS imported.
Also fixed some styles.

### DIFF
--- a/src/lib/components/footer/QFooter.svelte
+++ b/src/lib/components/footer/QFooter.svelte
@@ -2,12 +2,12 @@
   import { getContext } from "svelte";
   import type { QFooterProps } from "./props";
   import { createClasses, createStyles } from "$lib/utils/props";
-  import type { AppbarContext, LayoutContext } from "../layout/QLayout.svelte";
+  import type { LayoutContext } from "../layout/QLayout.svelte";
   import { useSize } from "$lib/composables/use-size";
 
   export let value: QFooterProps["value"] = true,
-    bordered: QFooterProps["bordered"] = false,
-    elevated: QFooterProps["elevated"] = false,
+    border: QFooterProps["border"] = false,
+    elevate: QFooterProps["elevate"] = false,
     height: QFooterProps["height"] = undefined,
     userClasses: QFooterProps["userClasses"] = undefined,
     userStyles: QFooterProps["userStyles"] = undefined;
@@ -15,13 +15,17 @@
 
   let ctx = getContext<LayoutContext | undefined>("layout");
 
-  $: heightStyle = $ctx && useSize(height).style;
-
-  $: classes = createClasses(["q-footer", $ctx && $ctx.footer?.fixed && "fixed", userClasses]);
+  $: classes = createClasses(
+    [border && "bordered", elevate && "elevated", $ctx?.footer?.fixed && "fixed"],
+    {
+      component: "q-footer",
+      userClasses,
+    }
+  );
 
   $: style = createStyles(
     {
-      "--footer-height": heightStyle,
+      height: !ctx && useSize(height).style,
     },
     userStyles
   );
--- a/src/lib/components/footer/index.scss
+++ b/src/lib/components/footer/index.scss
@@ -2,4 +2,27 @@
   width: 100%;
   height: var(--footer-height);
   transition: all var(--speed3);
+  background-color: var(--surface);
+  color: var(--on-surface);
+  padding: 0 1rem;
+
+  nav {
+    display: flex;
+    align-items: center;
+    justify-content: flex-start;
+    white-space: nowrap;
+    gap: 1rem;
+
+    > * {
+      margin: 0;
+    }
+  }
+
+  &--bordered {
+    border-top: solid 0.0625rem var(--outline);
+  }
+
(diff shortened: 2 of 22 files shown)


src/lib/components/footer/QFooter.svelte
```diff
@@ -2,26 +2,30 @@
...
   import type { QFooterProps } from "./props";
   import { createClasses, createStyles } from "$lib/utils/props";
-  import type { AppbarContext, LayoutContext } from "../layout/QLayout.svelte";
+  import type { LayoutContext } from "../layout/QLayout.svelte";
   import { useSize } from "$lib/composables/use-size";
 
   export let value: QFooterProps["value"] = true,
-    bordered: QFooterProps["bordered"] = false,
-    elevated: QFooterProps["elevated"] = false,
+    border: QFooterProps["border"] = false,
```
therealcoder1337, line 9: It appears these should be `bordered` and `elevated`, not sure why they were changed.

src/lib/components/footer/index.scss
```diff
@@ -2,4 +2,27 @@
   width: 100%;
   height: var(--footer-height);
   transition: all var(--speed3);
+  background-color: var(--surface);
+  color: var(--on-surface);
+  padding: 0 1rem;
+
+  nav {
```
therealcoder1337, line 9: Okay for now, but maybe we should use a class selector at some point.

src/lib/components/header/QHeader.svelte
```diff
@@ -0,0 +1,37 @@
...
+    console.warn("QHeader should be used inside QLayout");
+  }
+
+  $: classes = createClasses([$ctx?.header?.fixed && "fixed"], {
+    component: "q-header",
+    userClasses,
+  });
+
+  $: if ($ctx?.header !== undefined) {
+    if (userStyles?.includes("display: none")) {
```
therealcoder1337, line 27: You should be able to use the ternary operator here:
```ts
$ctx.header.display = userStyles?.includes("display: none") ? false : true;
```
Or even shorter:
```ts
$ctx.header.display = !userStyles?.includes("display: none");
```

src/lib/components/layout/index.scss
```diff
@@ -9,6 +9,8 @@
   --right-drawer-width: 300px;
 }
 
+$contentSlot: ":not(.q-header, .q-footer, .q-railbar, .q-drawer)";
```
therealcoder1337, line 12: Do we want to use `camelCase` or `kebab-case` for SCSS vars? So far I've used the latter, but I'm open to discuss it.

src/lib/components/layout/index.scss
```diff
@@ -149,29 +151,20 @@
...
-    &__right {
-      &--top {
-        border-top-right-radius: 16px;
-      }
-      &--bottom {
-        border-bottom-right-radius: 16px;
+  $horizontal: "left", "right";
+  $vertical: "top", "bottom";
+
+  @each $vPos in $vertical {
```
therealcoder1337, line 164: Same here, also `$verticalPos` or `$vertical-pos` would be a bit more descriptive. Not a blocker, just a minor comment.

src/lib/components/railbar/index.scss
```diff
@@ -0,0 +1,35 @@
...
+  justify-content: flex-start;
+  text-align: center;
+  vertical-align: middle;
+
+  border: 0;
+  border-radius: 0;
+  margin: 0;
+  padding: 0.5rem;
+
+  position: fixed;
```
therealcoder1337, line 14: Another minor thing: Why are there spaces? Also, the more important CSS properties should come first, so it's kinda hard to understand why `position` is below `border`.^^

src/lib/components/toolbar/QToolbar.svelte
```diff
@@ -1,41 +1,24 @@
...
-    } else {
-      $ctx.header.display = true;
-    }
-  }
+  });
 
   $: style = createStyles(
     {
-      "--header-height": heightStyle,
+      height: !userClasses?.includes("q-header") && useSize(height).style,
```
therealcoder1337, line 21: What could a falsy `height` evaluate to? Could it cause invalid CSS?

src/lib/utils/props.ts
```diff
@@ -1,14 +1,14 @@
...
-  styleObj: Record<string, string | number | null | undefined>,
+  styleObj: Record<string, string | number | boolean | null | undefined>,
   userStyles?: string
 ) {
   const stylesArray = Object.entries(styleObj);
   const toJoin: string[] = [];
 
   for (let [styleName, styleVal] of stylesArray) {
-    if (styleVal === undefined || styleVal === null) {
+    if (styleVal === undefined || styleVal === null || styleVal === false) {
```
therealcoder1337, line 11: In JS you check for `undefined` via `typeof`, here: `typeof styleVal === 'undefined'`. The others are correct.

therealcoder1337 (approved):
There are quite a few things to look at, but no blocks.
Great job! Seems like a *lot* of work.

PR #31, opened 2023-08-16
Refactors 16 aug

therealcoder1337:


src/lib/stores/QTheme.ts
```diff
@@ -45,7 +45,17 @@ interface IMaterialDynamicColorsThemeFormatted {
 type CSSDynamicColor =
   `${keyof IMaterialDynamicColorsThemeColorFormatted}-${keyof IMaterialDynamicColorsThemeFormatted}`;
 
+function extractColorFromCssVar(cssVar: string) {
+  const rootStyles = getComputedStyle(document.documentElement);
+  const varName = cssVar.replace(/var\(([a-z0-9-]+)\)/, "$1");
```
KilDesu, line 50: Wouldn't it be more efficient to use `cssVar.slice(4, cssVar.length - 1)`? We might want to avoid regex as much as possible

src/lib/stores/QTheme.ts
```diff
@@ -69,10 +79,9 @@ async function prepareThemeColors(from: string) {
   return themeColors;
 }
 
-const defaultThemeColors = await prepareThemeColors("#3499E7");
-
 function themeBuilder() {
-  const { subscribe, set, update } = writable(defaultThemeColors);
+  const { subscribe, set, update } = writable({} as Awaited<ReturnType<typeof prepareThemeColors>>);
```
KilDesu, line 83: Quite ugly solution indeed. We might really want to change this in the future. Could you open an issue?

KilDesu (approved):
Good job, please read my comments

PR #33, opened 2023-08-17
Refactors 17 aug

therealcoder1337:
This should clean up most TypeScript errors for users.

### DIFF
--- a/src/lib/composables/use-router-link.ts
+++ b/src/lib/composables/use-router-link.ts
@@ -4,9 +4,9 @@ import type { Page } from "@sveltejs/kit";
 export interface UseRouterLinkProps {
   href?: string;
   to?: string;
-  disable: boolean;
+  disable?: boolean;
   activeClass?: string;
-  replace: boolean;
+  replace?: boolean;
 }
 
 export const UseRouterLinkPropsDefaults: UseRouterLinkProps = {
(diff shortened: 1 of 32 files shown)


src/lib/composables/use-router-link.ts
```diff
@@ -27,13 +27,12 @@ export function isRouteActive(
 }
 
 export default function <T extends UseRouterLinkProps>(props: T) {
-  const hasLink = props.to !== undefined || props.href !== undefined;
-
+  const hasLink = [props.to, props.href].some((entry) => typeof entry !== "undefined");
```
KilDesu, line 30: Why `typeof entry !== "undefined"` and not `entry !== undefined`?

KilDesu (approved):
Good job, thnak you! I think we'll need to find a better solution than making all props optional. I'll open an issue about that.
I've written one comment, not very important though.

src/lib/composables/use-router-link.ts
therealcoder1337: Good question. Because it's one of the safest ways to check if something is `undefined`, even though it became a bit less needed with ES5. Still, in some contexts `undefined` could be [be reassigned](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined#description).
So I just prefer the [`typeof` check](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined#typeof_operator_and_undefined). But we could also agree on [`void 0`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined#void_operator_and_undefined) if you prefer that?

PR #34, opened 2023-08-18
feat: improve logo svg, change favicon, adjust home page.

therealcoder1337:
This logo edit aligns much better with a 32x32 px grid. I also removed some paths from where they shouldn't be.


therealcoder1337:
@KilDesu Hmm I was thinking if I should move the new logo more to the left, so the cocktail is centered (without the slice of orange). What do you think?

KilDesu (approved):

KilDesu:
> @KilDesu Hmm I was thinking if I should move the new logo's content more to the left, so the cocktail is centered (not considering the slice of orange). What do you think?

@therealcoder1337 Yes sure!

PR #36, opened 2023-08-22
Kil refactors 21aug

KilDesu:
Multiple refactors

### DIFF
--- a/src/lib/components/list/QItem.svelte
+++ b/src/lib/components/list/QItem.svelte
@@ -22,8 +22,8 @@
     userClasses: QItemProps["userClasses"] = undefined;
   export { userClasses as class };
 
-  let hasMultiplLines = writable(false);
-  setContext("hasMultipleLines", hasMultiplLines);
+  let hasMultipleLines = writable(false);
+  setContext("hasMultipleLines", hasMultipleLines);
 
   $: ({ hasLink, linkAttributes, linkClasses } = useRouterLink({
     href,
@@ -43,7 +43,7 @@
 
   $: classes = createClasses(
     [
-      $hasMultiplLines && "multiline",
+      $hasMultipleLines && "multiline",
       dense && "dense",
       (isActive || (hasLink && active)) && activeClass,
       (isActive || (hasLink && active)) && "active",
--- a/src/lib/components/list/QItem.svelte
+++ b/src/lib/components/list/QItem.svelte
@@ -18,6 +18,7 @@
     disable: QItemProps["disable"] = false,
     activeClass: QItemProps["activeClass"] = undefined,
     replace: QItemProps["replace"] = false,
+    noRipple: QItemProps["noRipple"] = false,
     userClasses: QItemProps["userClasses"] = undefined;
   export { userClasses as class };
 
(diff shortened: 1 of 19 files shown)


src/lib/components/list/QItem.svelte
```diff
@@ -33,23 +34,25 @@
...
-    hasLink && active && "q-item--active",
-    hasLink && active && activeClass,
-    isActive && "active",
-    isClickable && "wave",
-    linkClasses,
-    userClasses,
-  ]);
+  $: classes = createClasses(
+    [
+      $hasMultiplLines && "multiline",
```
therealcoder1337, line 45: Can we use full words for this please? E.g. `hasMultipleLines`

src/lib/components/list/QItemSection.svelte
```diff
@@ -15,14 +15,17 @@
...
+  $: classes = createClasses([type], {
+    component: "q-item__section",
+    userClasses,
+  });
 </script>
 
 <div class={classes} {...$$restProps}>
   {#if type === "content"}
-    {#if [$$slots.headline, $$slots.line2, $$slots.line2, $$slots.line3].some(Boolean) === false}
+    <!-- {#if [$$slots.headline, $$slots.line2, $$slots.line2, $$slots.line3].some(Boolean) === false}
```
therealcoder1337, line 26: Why is this commented out? Can we remove it?

src/lib/stores/Quaff.ts
```diff
@@ -10,30 +10,37 @@ function quaff() {
...
+          };
 
       let body = document.querySelector("body");
-
-      if (q.dark === true) {
-        body && body.classList.replace("light", "dark");
-      } else {
-        body && body.classList.replace("dark", "light");
+      if (body) {
+        body.classList.replace(`body--${mode.from}`, `body--${mode.to}`);
```
therealcoder1337, line 30: While I like the conciseness of it, the downside is you have to set a `body--light` class initially in your layout. This makes it somewhat fragile from an engineering standpoint. We should think if we can find another way, while keeping the code short. Also, I think in case the `q.dark` mismatches the class on the layout, it will be even more problematic and stop working completely.

src/lib/utils/props.ts
```diff
@@ -29,7 +29,7 @@ export function createStyles(
 interface CreateClassesOptions {
   component?: string;
   element?: string;
-  userClasses?: string;
+  userClasses?: string | null;
```
therealcoder1337, line 32: Why?

therealcoder1337 (approved):
Great work!

src/lib/utils/props.ts
KilDesu: Because I had an error `type string | undefined | null is not assignable to type string | undefined`. `string | undefined | null` is the type of the html "class" attribute

src/lib/components/list/QItemSection.svelte
KilDesu: I was not sure why it was here so I commented out in case I had to go back to it. I was right because I had to fix it in the last commits

src/lib/components/list/QItem.svelte
KilDesu: We sure can

src/lib/stores/Quaff.ts
KilDesu: Don't worry, I'm caching the mode inside cookies to be able to read it in server side.
If there's no cookie, the mode is already automatically set to "light" and a cookie is created on server side. So when this code runs, there's always a cookie

PR #39, opened 2023-08-23
Kil refactors aug23

KilDesu:


src/lib/components/layout/QLayout.svelte
```diff
@@ -1,4 +1,11 @@
 <script context="module" lang="ts">
+  import { createStyles, createClasses } from "$lib/utils";
+  import { setContext } from "svelte";
+  import type { Readable } from "svelte/motion";
```
therealcoder1337, line 4: 1. You probably want `svelte/store`, no?
2. Can we group `import type` statements above each other? I prefer to put them last, example:
```ts
import ...
import type ...
import type ...
```

src/lib/components/list/QItemSection.svelte
```diff
@@ -23,32 +23,32 @@
 
 <div class={classes} {...$$restProps}>
   {#if type === "content"}
-    {#if [$$slots.headline, $$slots.line2, $$slots.line2, $$slots.line3].every(Boolean) === false}
+    {#if [$$slots.headline, $$slots.line2, $$slots.line2, $$slots.line3].some(Boolean) === false}
```
therealcoder1337, line 26: Use `![$$slots.headline ...` instead of `... === false`.

src/lib/components/tooltip/props.ts
```diff
@@ -1,5 +1,4 @@
-import { NativePropsDefaults } from "$lib/utils/types";
-import type { NativeProps } from "$lib/utils/types";
+import { type NativeProps, NativePropsDefaults } from "$lib/utils";
```
therealcoder1337, line 1: Please split type and non-type imports into two statements.

src/lib/composables/use-align.ts
```diff
@@ -34,7 +34,7 @@ const alignMap = {
   stretch: "stretch",
 } as const;
 
-export default function useAlign(align: UseAlignOptions = "top left") {
+export function useAlign(align: UseAlignOptions = "top left") {
```
therealcoder1337, line 37: Why?

src/lib/css/theme/colors.module.scss
therealcoder1337, line 1: (How) is this file generated?

therealcoder1337 (approved):
Wow.

src/lib/composables/use-align.ts
KilDesu: Just for consistency, because other composables also are named exports

src/lib/css/theme/colors.module.scss
KilDesu: It was generated using MD3's [theme builder](https://m3.material.io/theme-builder#/custom).
We might want to find a solution to automatically change it because doing manually each time it's generated will quickly become a hassle.

PR #43, opened 2023-09-09
feat: refactor q-input and q-select.

### LINKED ISSUE #9: QInput and QSelect label animation via CSS
The elements should be refactored to rely only on CSS for their label animations.
Right now we're using JS for the animation and one tester reported problems with that. There are also minor UI glitches.

therealcoder1337:
Fixes #9.

Kinda want to have this merged, as it contains many changes.
Known issues:

- `QSelect`: Menu opens below the hint, if a hint is specified
- `QSelect`: Sometimes there's a slight vertical shift of the label upon first focusing the select

It would be great if you could also try this out in your browser for a bit.


src/lib/components/input/QInput.svelte
```diff
@@ -13,89 +16,70 @@
...
+  class:q-field--has-border={outlined || rounded}
+  class:q-field--dense={dense}
+  class:q-field--active={active}
+  class:q-field--focus={focus}
+  class:q-field--label={label}
+  class:q-field--slot-append={$$slots.append}
+  class:q-field--slot-prepend={$$slots.prepend}
+  class:q-field--disable={disable}
+  class:q-field--error={error}
+  style="--slot-prepend-width: {slotPrependWidth}px"
```
KilDesu, line 39: Can change this to `style:--slot-prepend-width="{slotPrependWidth}px"`

src/lib/components/input/input-mixins.scss
```diff
@@ -0,0 +1,52 @@
...
+    0% 100%,
+    100% 100%,
+    100% $clip-height,
+    calc(100% - #{$clip-width}) $clip-height,
+    calc(100% - #{$clip-width}) 0%
+  );
+}
+
+@mixin border-substitute($variant: "outlined") {
+  $border-radius: if($variant == "outlined", 4px, calc(var(--wrapper-height) / 2));
```
KilDesu, line 23: Maybe use `rem` instead of `px`?

src/lib/components/input/input-mixins.scss
```diff
@@ -0,0 +1,52 @@
...
+    border-top-color: var(--decorator-color);
+    border-radius: $border-radius;
+  }
+
+  &::before {
+    left: calc(var(--border-width) * -1);
+    border-top-right-radius: 0px;
+    border-bottom-right-radius: 0px;
+    $clip-height: 2px;
+    $clip-width: if($variant == "outlined", 15px, 5px);
```
KilDesu, line 43: See last comment

src/lib/components/select/QSelect.svelte
```diff
@@ -153,47 +104,86 @@
...
+  class:q-field--has-border={outlined || rounded}
+  class:q-field--dense={dense}
+  class:q-field--active={active}
+  class:q-field--focus={focus}
+  class:q-field--label={label}
+  class:q-field--slot-append={$$slots.append}
+  class:q-field--slot-prepend={$$slots.prepend}
+  class:q-field--disable={disable}
+  class:q-field--error={error}
+  style="--slot-prepend-width: {slotPrependWidth}px"
```
KilDesu, line 122: See my first comment

src/lib/components/select/QSelect.svelte
```diff
@@ -153,47 +104,86 @@
...
+        <div class="q-field__slot-prepend" bind:clientWidth={slotPrependWidth}>
+          <slot name="prepend" />
+        </div>
+      {/if}
+
+      <input
+        class="q-field__input"
+        bind:value
+        placeholder=""
+        on:focus={(e) => handleFocus(e.currentTarget)}
```
KilDesu, line 143: Prefer a function that takes the event as the parameter and take the target inside it. You could do it like this:
```js
function handleFocus({ currentTarget }: FocusEvent) { // Or whatever type that is
  // ...
}
```
So you can do `on:focus={handleFocus}`

src/lib/components/select/QSelect.svelte
```diff
@@ -153,47 +104,86 @@
...
+        on:blur={(e) => handleBlur(e.currentTarget)}
```
KilDesu, line 144: idem

src/lib/components/select/QSelect.svelte
```diff
@@ -153,47 +104,86 @@
...
+        on:mousedown={() => handleMousedown()}
```
KilDesu, line 145: `on:mousedown={handleMousedown}`

KilDesu (approved):
Some comments but not very break changing overall, thanks for your hard work!

src/lib/components/input/input-mixins.scss
therealcoder1337: Thanks, this file shouldn't exist anyway, since the new variant is called `field-mixins.scss` and located in another folder.

therealcoder1337:
Thanks for the review! Most issues are fixed now, I could also resolve the `hint` issue for `QSelect`.

PR #45, opened 2023-09-09
Qtabs refactor

KilDesu:


src/lib/components/tabs/QTab.svelte
```diff
@@ -3,87 +3,79 @@
...
-
-  function setActive(el: HTMLElement) {
-    let rect;
-    if (variant === "primary") {
-      rect = (el.firstElementChild as HTMLElement).getBoundingClientRect();
-    } else {
-      rect = el.getBoundingClientRect();
-    }
+  function setActive(el: QTab) {
+    let store = get(qTabStore);
```
therealcoder1337, line 53: Can this be a `const`?

src/lib/components/tabs/QTabs.svelte
```diff
@@ -1,67 +1,103 @@
...
 
-  /** Counting QTab children number */
-  let QTabCount = 0;
-  setContext("QTabCount", {
-    index: () => {
-      QTabCount += 1;
-      return QTabCount;
-    },
-  });
+  /* const indicatorWidth = derived(qTabStore, ($store) => {
```
therealcoder1337, line 35: Please remove comments like this, unless they're important.

src/lib/components/tabs/QTabs.svelte
```diff
@@ -1,67 +1,103 @@
...
+  $: value = $qTabStore.value;
 
-  $: value = $activeTabStore.name;
+  setContext("qTabStore", qTabStore);
 
-  $: classes = createClasses([variant, round && "rounded", hidden && "hidden-indicator"], {
-    component: "q-tabs",
-    userClasses,
-  });
+  $: if ($qTabStore.activeEl !== null) {
```
therealcoder1337, line 50: Is it possible to remove the `!== null`?

src/lib/components/tabs/QTabs.svelte
```diff
@@ -1,67 +1,103 @@
...
 
-  let indicatorWidth = derived(activeTabStore, ($activeTabStore) => {
-    return variant === "primary"
-      ? `calc(${$activeTabStore.size}px + 8px)`
-      : `${$activeTabStore.size}px`;
-  });
+    const child = variant === "primary" ? activeEl.firstElementChild as HTMLDivElement : { offsetLeft: 0, offsetWidth: 0 }
+    const previousChild = variant === "primary" ? previousEl?.firstElementChild as HTMLDivElement : { offsetLeft: 0, offsetWidth: 0 }
+
+    if (previousEl === null) {
```
therealcoder1337, line 63: Is it possible to use just `!previousEl`?

src/lib/components/tabs/QTabs.svelte
```diff
@@ -1,67 +1,103 @@
...
+      qTabs.style.setProperty("--indicator-position", `${position}px`);
+    } else {
+      // Position indicator on tab change
+      const comparePositions = movementDirection(previousEl, activeEl);
+
+      let transitionSize;
+      if (comparePositions === "next") {
+        // New tab is after the previous one
+        transitionSize =
+          storeVariant === "vertical"
```
therealcoder1337, line 75: This seems to be the same algorithm as in the `else` branch, only with different variable order, if I'm not mistaken. If so, this should be refactored into a function that takes two parameters.

src/lib/components/tabs/QTabs.svelte
```diff
@@ -1,67 +1,103 @@
...
-    {
-      "--tab-count": QTabCount || 1,
-      "--indicator-size": $indicatorWidth,
-      "--active-tab-index": $activeTabStore.index - 1,
-      "--indicator-position": $activeTabStore.position,
-    },
-    userStyles
-  );
+      setTimeout(() => {
+        qTabs.style.setProperty("--indicator-position", `${position}px`);
```
therealcoder1337, line 90: You could use an enum or hard coded constants instead of strings for these strings of CSS variables, so you won't have typos somewhere.

src/lib/components/tabs/index.scss
```diff
@@ -70,38 +69,48 @@
...
-      top: var(--indicator-position);
-      right: 0;
-      bottom: auto;
       left: auto;
-      width: 2px;
-      height: var(--indicator-size);
+      right: 0;
+      top: 0;
+      bottom: 0;
+
```
therealcoder1337, line 76: Do we really want these empty lines?

therealcoder1337 (approved):
LGTM, thanks. :) Some parts were a bit hard to review because of the constant switch between deletions and insertions.

PR #47, opened 2023-09-13
13 sep work

therealcoder1337:
- Worked on QProgress and docs
- QField refactor to prevent field growth on focus

### DIFF
--- a/src/lib/components/dialog/QDialog.svelte
+++ b/src/lib/components/dialog/QDialog.svelte
@@ -1,6 +1,5 @@
 <script lang="ts">
   import { clickOutsideDialog } from "$lib/helpers";
-  import { createClasses } from "$lib/utils";
   import { createEventDispatcher } from "svelte";
   import { QBtn } from "$lib";
   import type { QDialogProps } from "./props";
@@ -34,16 +33,6 @@
   }
 
   $: canHideOnClickOutside = value && !persistent;
-
-  $: positionClass = ["top", "right", "bottom", "left"].includes(position!) ? position : undefined;
-
-  $: classes = createClasses(
-    [value && "active", positionClass, modal && "modal", fullscreen && "max"],
-    {
-      component: "q-dialog",
-      userClasses,
-    }
-  );
 
   export function hide() {
     value = false;
@@ -98,7 +87,15 @@
 
 <dialog
   use:clickOutsideDialog={handleClickHide}
-  class={classes}
+  class="q-dialog {userClasses}"
+  class:q-dialog--active={value}
+  class:q-dialog--position-default={position === "default" || !position}
+  class:q-dialog--position-top={position === "top"}
+  class:q-dialog--position-right={position === "right"}
+  class:q-dialog--position-bottom={position === "bottom"}
+  class:q-dialog--position-left={position === "left"}
+  class:q-dialog--modal={modal}
+  class:q-dialog--fullscreen={fullscreen}
   {...$$restProps}
   bind:this={dialogElement}
   on:cancel={handleKeyboardHide}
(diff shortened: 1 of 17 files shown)


src/lib/components/progress/QCircularProgress.svelte
```diff
@@ -18,23 +18,34 @@
...
     userClasses,
   });
+
+  $: circumference = 2 * Math.PI * 20;
+
+  $: progressOffset = ((100 - value) / 100) * circumference;
 </script>
 
 <svg
   class={classes}
```
KilDesu, line 28: Possible to refactor to class directives?

src/lib/components/progress/QCircularProgress.svelte
```diff
@@ -18,23 +18,34 @@
     quaffClasses: [color && `text-${color}`],
     userClasses,
   });
+
+  $: circumference = 2 * Math.PI * 20;
```
KilDesu, line 22: This shouldn't be a reactive value, it should be a `const`. Or maybe you wanted it to change with `size`?

KilDesu (approved):
Some comments but nothing much.
Maybe you can refactor classes and styles for the lunear progress as well, would be great. If you can't, we'll be doing it in other prs.
Thanks for your work!

src/lib/components/dialog/QDialog.svelte
```diff
@@ -1,77 +1,58 @@
...
+    userClasses: QDialogProps["userClasses"] = undefined,
+    value: QDialogProps["value"] = false;
   export { userClasses as class };
 
   const emit = createEventDispatcher();
-  let dialogElement: HTMLDialogElement;
+  let dialogElement: HTMLDialogElement | undefined;
 
-  let opened = false;
+  const DURATION_TOGGLE = 150;
```
KilDesu, line 21: Can we change this to `TOGGLE_DURATION` or something different like `ANIMATION_DURATION`?

src/lib/components/dialog/QDialog.svelte
```diff
@@ -1,77 +1,58 @@
...
-      }, 250);
+        dialogElement!.style.display = "none";
+      }, DURATION_TOGGLE);
     }
   }
+
+  $: canHideOnClickOutside = value && !persistent;
+
+  export function hide() {
+    value = false;
```
KilDesu, line 38: I think `value` should be set to false only if it's true in the first place to avoid running unnecessary svelte reactivity code.
Maybe, just in case, consider changing this to
```js
if(value) {
  value = false
}
```

src/lib/components/dialog/QDialog.svelte
```diff
@@ -1,77 +1,58 @@
...
+  }
+
   export function show() {
-    if (dialogElement && !dialogElement.open) {
-      modal ? dialogElement.showModal() : dialogElement.show();
-      opened = true;
-      dialogElement.style.display = "block";
-    }
+    value = true;
```
KilDesu, line 42: Same as last comment but checking for `!value`

src/lib/components/progress/QLinearProgress.svelte
```diff
@@ -13,23 +13,25 @@
...
     from === "right"
       ? `clip-path: polygon(100% 0%, 100% 100%, ${100 - value}% 100%, ${100 - value}% 0%);`
       : `clip-path: polygon(0% 0%, 0% 100%, ${value}% 100%, ${value}% 0%);`;
 </script>
 
-<div class={containerClasses} style={containerStyle} {...$$restProps}>
-  <div class={progressClasses} style={progressStyle} />
+<div
+  class="q-linear-progress {userClasses}"
+  style={containerStyle}
```
KilDesu, line 25: Probably better to have `style:border-radius={rounded ? "12px" : "0"}` instead of using `createStyles`. That would also allow us to remove the `userStyles` prop

KilDesu (approved):
Good work, as always!

PR #48, opened 2023-09-17
docgen overhaul

therealcoder1337:
- docgen is now multi threaded
- It is ran automatically via a Vite plugin on both dev server start and build
- Hot reload is also supported; when you change a `props.ts`, docgen runs again
- Little downside: we have to use `spawn()` because I couldn't get node Workers supported using Vite


docgen/props/run.ts
```diff
@@ -1,50 +1,80 @@
...
+  // sort alphabetically
+  const allTypesSorted = Object.keys(allTypesRaw)
+    .sort()
+    .reduce(
+      (carry, key) => {
+        carry[key] = allTypesRaw[key];
+        return carry;
+      },
+      {} as Record<string, string>
+    );
```
KilDesu, line 38: Please don't use weird `reduce` functions for that :
```js
const allTypesSorted = Object.fromEntries(
  Object.entries(allTypesRaw).sort((a, b) => a[0].localeCompare(b[0]))
)
```

If you want to ignore accents and case, use `a.localeCompare(b, "en", { sensitivity: "base" })`

src/dev/docgen-plugin.ts
```diff
@@ -0,0 +1,68 @@
...
+    handleHotUpdate({ file, server }: { file: string; server: ViteDevServer }) {
+      const now = new Date().getTime();
+
+      if (lastUpdated.file === file && now - lastUpdated.at < 500) {
+        // throttle calls, in case of formatters etc.
+        return;
+      }
+
+      lastUpdated.file = file;
+      lastUpdated.at = now;
```
KilDesu, line 56: The if statement will always run, no? Because now will be the same inside the `if` and `lastUpdated.at` no matter what (as you don't rerun `new Date.now()`)

KilDesu (approved):
Great job again! Only few comments

docgen/props/WorkerManager.ts
```diff
@@ -0,0 +1,90 @@
...
+  private maxWorkers: number;
+  private workers: ExtendedWorker[];
+  private tasks: WorkerTask[];
+  private types: Record<string, string>;
+
+  constructor(workerPath: string, tasks: WorkerTask[]) {
+    super();
+    this.workerPath = workerPath;
+    const availableWorkers = os.availableParallelism();
+    this.maxWorkers = availableWorkers >= 6 ? availableWorkers - 2 : availableWorkers - 1;
```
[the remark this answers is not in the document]
therealcoder1337: It's harder to read and will def. not have any noticeable performance impact.

docgen/props/run.ts
therealcoder1337: Thanks, will consider.

src/dev/docgen-plugin.ts
therealcoder1337: Don't understand what you mean here. It should only return when now minus the last updated time are below 500ms. Since `lastUpdated.at` gets only assigned after the `if`, it should be fine.

PR #50, opened 2023-09-17
Qtooltip refactor

KilDesu:
Refactored ripple so it doesn't interfere with QTooltip. Also refactored QTooltip to use dynamic classes and fixed cursor.
There's still a "problem" of text not being selectable inside the tooltip but it's maybe better like this, buttons still seem to be clickable.


therealcoder1337 (approved):
LGTM, thanks!

PR #51, opened 2023-09-17
Code fixes

therealcoder1337:
This fixes most Prettier and ESLint errors. One ESLint error remains, maybe you can look into it @KilDesu:

```shell
$ npx eslint .

quaff/src/lib/components/list/QItemSection.svelte
  10:6  error  'ctx' is defined but never used  @typescript-eslint/no-unused-vars
````
Notes from the commit message:

- removed unused "getAllProperties" from parse-interface
- refactored QCodeBlock to support svelte and include styles
- QDrawer: lots of currently unused props were removed
- QFooter: wrapped inside an if statement to actually use "value"
- removed many (apparently) unused styles
- props should never have the same variable name as their type, so I had
to change QComponentDocs to prop name componentDocs

I'll likely add a few more changes to this branch.

### CODE REVIEW

**Comment by therealcoder1337** [comment]
@KilDesu: I've also added automatic type check and lint for PRs now. After you've fixed the last ESLint error in this branch, the checks should pass.

**Comment by KilDesu** [comment]
> @KilDesu: I've also added automatic type check and lint for PRs now. After you've fixed the last ESLint error in this branch, the checks should pass.

Just checked the error, it is unfounded because `ctx` is used: it's set just few lines below. I think eslint is expecting it to be used at least once in "read" but as this is a writable store, it's only used in "write" so the parent can use it in "read". This error shouldn't be here.

File: src/lib/components/radio/props.ts [hunk 1]
```diff
@@ -5,7 +5,7 @@ import type { NativeProps } from "$lib/utils";
 export interface QRadioProps extends NativeProps, HTMLAttributes<HTMLLabelElement> {
   value: string;
   label?: string;
-  selected?: any;
+  selected?: unknown;
```
KilDesu, line 8: What's this prop for, and why is it unknow at compile time?

src/lib/composables/useSize.ts
```diff
@@ -8,13 +8,13 @@ interface UseSize {
   style?: string;
 }
 
-export function useSize(sizeProp: any): UseSize {
+export function useSize(sizeProp: unknown): UseSize {
```
KilDesu, line 11: Not sure I agree with this one... The sizeProp should be `any` type, not an `unknown` type imho

src/lib/utils/string.ts
```diff
@@ -39,6 +39,8 @@ function convertToKebabSnakeCase(str: string, caseType: keyof typeof cases) {
 }
 
 export function convertCase(str: string, fromCase: keyof typeof cases, toCase: keyof typeof cases) {
+  const uncap = uncapitalize(str);
+
```
KilDesu, line 43: Don't put it here, it should be inside the "pascal" arm case as it shouldn't run for all case arms

src/lib/utils/props.ts
KilDesu, line 1: This file should soon be deleted as we want to get rid of the `createClasses` and `createStyles` functions but it's not urgent atm

KilDesu (approved):
Good job! I have some comments so please consider them before merging

src/lib/utils/string.ts
therealcoder1337: It's because of ESLint, the function needs to be refactored anyway.

src/lib/composables/useSize.ts
therealcoder1337: No, `any` gives an error with our config. This is good, because `unknown` should be preferred over `any`.

therealcoder1337: Although it would be even better to have an explicit type!

src/lib/components/radio/props.ts
therealcoder1337: Short story: Because I couldn't find the proper type. But it should be addressed in the future.

src/lib/components/card/QCard.svelte
```diff
@@ -18,7 +18,7 @@
     : "surface-variant";
 </script>
 
-<article 
+<article
```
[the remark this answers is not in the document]
therealcoder1337: ?

PR #55, opened 2023-09-19
Snippets

therealcoder1337:
Same thing as with props, and a few refactors.


docgen/helpers/getComponentDirs.ts
```diff
@@ -0,0 +1,6 @@
+import { readdir } from "fs/promises";
+
+export default async function getComponentDirs(componentsSir: string) {
+  const dirents = await readdir(componentsSir, { withFileTypes: true });
```
KilDesu, line 4: You probably wanted to write `componentsDir`?

docgen/snippets/updateAllSnippets.ts
```diff
@@ -0,0 +1,21 @@
...
+
+const dirname = path.dirname(fileURLToPath(import.meta.url));
+const rootDir = path.resolve(dirname, "../../src/routes/components");
+
+export default async function updateAllSnippets() {
+  const componentsToIgnore = ["layout"];
+
+  const componentDirs = (await getComponentDirs(rootDir)).filter(
+    (dir) => !componentsToIgnore.includes(dir)
+  );
```
KilDesu, line 14:
```js
const componentsDirs = (await getComponentsDirs(rootDir)).filter(!componentsToIgnore.includes)
```

src/dev/docgenPlugin.ts
```diff
@@ -40,12 +47,19 @@ function docgenPlugin(): Plugin {
     name: "docgen-plugin",
     async configResolved(config) {
       config.logger.info(DOCGEN_LOG_MESSAGE);
-      await runDocGenProps();
+      const p1 = runDocGenProps();
+      const p2 = updateAllSnippets();
+      await Promise.all([p1, p2]);
```
KilDesu, line 52: Do we really need two variables for that, as I don't see them reused somewhere else? Can't we just pass the functions to `Promise.all` directly?

KilDesu (approved):
I have some comments, but otherwise, lgtm!

src/dev/docgenPlugin.ts
therealcoder1337: Yeah we could. I find this slightly more readable, but will change it.

PR #64, opened 2024-05-16
Svelte 5 refactors coder

therealcoder1337:
## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

## What's new?

> List the changes

- Refactored more components to Svelte 5

## Screenshots

N/A

N/A

### DIFF
--- a/src/lib/components/toggle/QToggle.svelte
+++ b/src/lib/components/toggle/QToggle.svelte
@@ -6,12 +6,12 @@
     icon = undefined,
     label = undefined,
     leftLabel = undefined,
-    disable = undefined,
+    disable = false,
     ...props
   }: QToggleProps = $props();
 
   function toggle() {
-    if (disable !== true) {
+    if (!disable) {
       value = !value;
     }
   }
@@ -33,7 +33,7 @@
   aria-disabled={disable || undefined}
   role="switch"
   aria-checked={value}
-  tabindex="0"
+  tabindex={disable ? -1 : 0}
 >
   <label class="q-toggle__inner" class:q-toggle__inner--icon={icon}>
     <input bind:checked={value} type="checkbox" disabled={disable || undefined} />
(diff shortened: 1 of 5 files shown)


src/lib/components/toggle/QToggle.svelte
```diff
@@ -1,31 +1,39 @@
...
-    leftLabel: QToggleProps["leftLabel"] = undefined,
-    disable: QToggleProps["disable"] = undefined,
-    userClasses: QToggleProps["userClasses"] = "";
-  export { userClasses as class };
+  let {
+    value = $bindable(),
+    icon = undefined,
+    label = undefined,
+    leftLabel = undefined,
+    disable = undefined,
```
KilDesu, line 9: Should be `false` by default imo

src/lib/components/toggle/QToggle.svelte
```diff
@@ -1,31 +1,39 @@
...
+    ...props
+  }: QToggleProps = $props();
 
   function toggle() {
     if (disable !== true) {
```
KilDesu, line 14: Then you can write `if (!disable) {`

KilDesu: Also, why the if statement?

src/lib/components/toggle/QToggle.svelte
```diff
@@ -1,31 +1,39 @@
...
       value = !value;
     }
   }
+
+  Q.classes("q-toggle", {
+    classes: [props.class],
```
KilDesu, line 20: Same here, `bemClasses` before `classes`?

src/lib/components/toggle/QToggle.svelte
therealcoder1337: It wasn't changed as you can see. The `if` is probably to only reassign when the condition is met, instead of running every time regardless.

KilDesu (approved):

Issue #61, opened 2024-05-14
Svelte 5 support per component


KilDesu:
- [x] QAvatar
- [x] QBreadcrumbs
- [x] QBtn
- [x] QCard
- [x] QCheckbox
- [x] QChip
- [x] QCodeBlock
- [x] QDialog
- [x] QDrawer
- [x] QFooter
- [x] QHeader - Will mostly need API change
- [x] QIcon
- [x] QInput
- [x] QLayout
- [x] QList
- [x] QCircularProgress
- [x] QLinearProgress
- [x] QRadio
- [x] QRailbar
- [x] QSelect
- [x] QSeparator
- [x] QTable
- [ ] QTabs
- [x] QToggle
- [ ] QToolbar - Only partially, will mostly need API change
- [ ] QTooltip

PR #66, opened 2024-05-18
Qchip refactor

KilDesu:
## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes
- [ ] No

## What's new?

> List the changes

- Migration to Svelte 5
- Followed MD3 API rather than Quasar's
- Some minor changes in files "unrelated" to QChip

Important to note that for the `input` design, I omitted the "selected" state like we have on the `filter` design.
I don't see any use case for that.

I sometimes find MD3 specs kind of inconsistent, which makes decisions and refactors a bit harder sometimes.

## Screenshots

image

Related to #61


src/lib/css/mixins/_design.scss
```diff
@@ -45,24 +46,25 @@
 }
 
 @mixin disabled {
-  cursor: not-allowed;
-  box-shadow: none;
+  cursor: not-allowed !important;
```
therealcoder1337, line 49: Is `!important` really necessary? Also for the others following.

therealcoder1337 (approved):
Awesome work! Let me know when you considered the comments and I'll merge.

src/lib/css/mixins/_design.scss
KilDesu: I think so, those styles are important and the specificity is pretty low so easily overwritten

src/lib/components/avatar/QAvatar.scss
```diff
@@ -38,18 +38,17 @@
...
+  "lg": 3,
+  "xl": 4,
 );
 
 .q-avatar {
   @each $size, $val in $sizes {
     &--#{$size} {
-      height: 2rem + ($val - 1) * 0.5rem;
-      width: 2rem + ($val - 1) * 0.5rem;
+      --size: calc(2rem + #{$val} * 0.5rem);
```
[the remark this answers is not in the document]
KilDesu: Not sure that's really necessary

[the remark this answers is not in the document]
therealcoder1337: In the future, we should be clear with such things. No problem now.

src/lib/css/mixins/_design.scss
therealcoder1337: It's kind of a "last resort". Usually you should not use `!important`.

PR #72, opened 2024-05-19
refactor(QTable): migrate to Svelte 5.

therealcoder1337:
## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

## What's new?

> List the changes

- Migrated QTable to Svelte 5

## Screenshots

N/A

N/A

### DIFF
--- a/src/lib/components/table/QTable.svelte
+++ b/src/lib/components/table/QTable.svelte
@@ -2,45 +2,45 @@
   import { QIcon, QSelect, QBtn } from "$lib";
   import type { QTableProps, QTableColumn, QTableRow, QTableSort } from "./props";
 
-  export let columns: QTableProps["columns"] = [],
-    rows: QTableProps["rows"] = [],
-    flat: QTableProps["flat"] = undefined,
-    bordered: QTableProps["bordered"] = undefined,
-    dense: QTableProps["dense"] = false,
-    userClasses: QTableProps["userClasses"] = "";
-  export { userClasses as class };
+  let {
+    columns = [],
+    rows = [],
+    flat,
+    bordered,
+    dense = false,
+    bodyCell,
+    ...props
+  }: QTableProps = $props();
 
   function getField(fieldRaw: QTableColumn["field"], row: QTableRow): string | number {
     return typeof fieldRaw === "function" ? fieldRaw(row) : row[fieldRaw];
   }
 
-  let page = 1;
-  let rowsPerPage = 5;
-  let rowsPerPageOptions = [5, 10, 25, 50].map((e) => ({
-    label: e.toString(),
-    value: e.toString(),
-  }));
+  let page = $state(1);
+  let rowsPerPage = $state(5);
+  let rowsPerPageOptions = $state(
+    [5, 10, 25, 50].map((e) => ({
+      label: e.toString(),
+      value: e.toString(),
+    }))
+  );
+  let sort: QTableSort = $state(null);
 
-  let sort: QTableSort = null;
-  let rowsSorted: QTableRow[] = rows;
-  let rowsPaginated: QTableRow[] = [];
-  let numberFrom: number = 1;
-  let numberTo: number = 5;
-  let numberOf: number = rows.length;
+  const numberFrom: number = $derived(rowsPerPage * page - rowsPerPage + 1);
+  const numberTo: number = $derived(
+    rowsPerPage * page > rows.length ? rows.length : rowsPerPage * page
+  );
+  const numberOf: number = $derived(rows.length);
 
-  $: numberFrom = rowsPerPage * page - rowsPerPage + 1;
-  $: numberTo = rowsPerPage * page > rows.length ? rows.length : rowsPerPage * page;
-  $: numberOf = rows.length;
-  $: rowsPaginated = rowsSorted.slice(numberFrom - 1, numberTo);
-  $: {
+  $effect(() => {
     if (rowsPerPage * (page - 1) >= rows.length) {
       page = 1;
     }
-  }
+  });
 
-  $: {
+  const rowsSorted: QTableRow[] = $derived.by(() => {
     if (sort) {
-      rowsSorted = structuredClone(rows).sort((a: QTableRow, b: QTableRow) => {
+      return structuredClone(rows).sort((a: QTableRow, b: QTableRow) => {
         const valA = getField(sort!.columnField, a);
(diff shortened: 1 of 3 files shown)


src/lib/components/table/QTable.svelte
```diff
@@ -2,45 +2,45 @@
...
-    dense: QTableProps["dense"] = false,
-    userClasses: QTableProps["userClasses"] = "";
-  export { userClasses as class };
+  let {
+    columns = [],
+    rows = [],
+    flat = undefined,
+    bordered = undefined,
+    dense = false,
+    bodyCell = undefined,
```
KilDesu, line 11: You don't need `flat = undefined`, `flat` is enough

src/lib/components/table/QTable.svelte
```diff
@@ -2,45 +2,45 @@
...
+      value: e.toString(),
+    }))
+  );
+  let sort: QTableSort = $state(null);
+
+  const numberFrom: number = $derived(rowsPerPage * page - rowsPerPage + 1);
+  const numberTo: number = $derived(
+    rowsPerPage * page > rows.length ? rows.length : rowsPerPage * page
+  );
+  const numberOf: number = $derived(rows.length);
```
KilDesu, line 33: Do you really need to write the type explicitly for those 3?

src/lib/components/table/QTable.svelte
```diff
@@ -2,45 +2,45 @@
...
       page = 1;
     }
-  }
+  });
 
-  $: {
+  const rowsSorted: QTableRow[] = $derived.by(() => {
     if (sort) {
-      rowsSorted = structuredClone(rows).sort((a: QTableRow, b: QTableRow) => {
+      return structuredClone(rows).sort((a: QTableRow, b: QTableRow) => {
```
KilDesu, line 43: Use `rows.toSorted` instead of `structuredClose(rows).sort`.
I even think that `sort` would return `undefined` as it mutates the array, so it doesn't return it.
`rows.toSorted` will clone `rows` and return the sorted value.

src/lib/components/table/QTable.svelte
```diff
@@ -96,19 +98,23 @@
...
 
-<div class="q-table" {...$$restProps}>
-  <table
-    class="q-table__table {userClasses}"
-    class:q-table--flat={flat}
-    class:q-table--bordered={bordered}
-    class:q-table--dense={dense}
-  >
+<div {...props} class="q-table">
+  <table class="q-table__table" {...Q.classes}>
```
KilDesu, line 113: Do we really need to wrap the table in a div?

src/lib/components/table/QTable.svelte
```diff
@@ -156,14 +164,14 @@
...
       design="flat"
       disabled={page === 1}
-      on:click={() => (page = page - 1)}
+      onclick={() => (page = page - 1)}
     />
     <QBtn
       icon="chevron_right"
       size="sm"
       design="flat"
       disabled={page * rowsPerPage >= rows.length}
```
KilDesu, line 173: If we have this, why do we have the effect on like 35? Also, why is it `page - 1` on line 35 and `page` here?

src/lib/components/table/QTable.svelte
```diff
@@ -156,14 +164,14 @@
       size="sm"
       design="flat"
       disabled={page === 1}
-      on:click={() => (page = page - 1)}
+      onclick={() => (page = page - 1)}
```
KilDesu, line 167: Why not write `page--`?

src/lib/components/table/QTable.svelte
```diff
@@ -156,14 +164,14 @@
...
     />
     <QBtn
       icon="chevron_right"
       size="sm"
       design="flat"
       disabled={page * rowsPerPage >= rows.length}
-      on:click={() => (page = page + 1)}
+      onclick={() => (page = page + 1)}
```
KilDesu, line 174: Same, why not write `page++`?

src/lib/components/table/props.ts
```diff
@@ -51,4 +52,10 @@ export interface QTableProps extends NativeProps, HTMLAttributes<HTMLDivElement>
    * @default false
    */
   dense?: boolean;
+
+  /**
+   * Optionally pass a snippet to render each table cell.
+   * @default undefined
+   */
+  bodyCell?: Snippet<[{ column: QTableColumn; row: QTableRow; style: string }]>;
```
KilDesu, line 60: Not sure you need to wrap the object in an array in this type, I think only the object would work. Could you test if it works?

KilDesu (changes requested):
Good work! Some minor changes but nothing too important. Thanks!

src/lib/components/table/QTable.svelte
therealcoder1337: Yeah we have many of these unnecessary default values, also Docs sometimes differ from the actual value. We should discuss a general approach.

src/lib/components/table/QTable.svelte
therealcoder1337: No, but now it's sure the expressions actually return the correct type - at least from a TS perspective.

src/lib/components/table/QTable.svelte
therealcoder1337: It's a good question at to why we've (I've) used `structuredClone` in the first place, since a deep clone doesn't seem necessary now. Maybe the idea was to prevent the user from accidentally changing any objects using the field function.
You're right that without a clone, `toSorted` would be the better choice. Since we're however cloning, `.sorted` seems faster, because it doesn't create an additional copy.
Not part of the Svelte 5 refactor, so I'm not risking any breaking change.

src/lib/components/table/QTable.svelte
therealcoder1337: Yes, what's the alternative? There's a row selection at the bottom.

src/lib/components/table/QTable.svelte
therealcoder1337: WDYM? This one just decreases the page number by subtracting `1`.

src/lib/components/table/props.ts
therealcoder1337: Tested, that's why I used an array in the first pace. 😄

PR #74, opened 2024-05-20
refactor(QContext): change API to make it easier to use

KilDesu:
## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

## What's new?

> List the changes

- Improve `QContext` API to avoid unnecessary function calls
- Document the `QContext` class and methods
- Fix some unnecessary uses of `QContext`
- Change QList to use the new API

## Screenshots

N/A

N/A

### DIFF
--- a/src/lib/classes/QContext.svelte.ts
+++ b/src/lib/classes/QContext.svelte.ts
@@ -2,7 +2,7 @@ import { getContext, setContext } from "svelte";
 
 /**
  * This class allows to manipulate reactive contexts more easily.
- * It avoid having to pass a Svelte store down the components but rather use runes to keep the context reactive.
+ * It avoids having to pass a Svelte store down the components but rather use runes to keep the context reactive.
  *
  * This class should be used when the context has to be modified from a child component. Otherwise, using svelte's context API should be enough.
  */
--- a/src/lib/classes/QContext.svelte.ts
+++ b/src/lib/classes/QContext.svelte.ts
@@ -3,7 +3,7 @@ import { getContext, setContext } from "svelte";
 /**
  * This class allows to manipulate reactive contexts more easily.
  * It avoid having to pass a Svelte store down the components but rather use runes to keep the context reactive.
- * 
+ *
  * This class should be used when the context has to be modified from a child component. Otherwise, using svelte's context API should be enough.
  */
 export default class QContext<T> {
@@ -23,7 +23,7 @@ export default class QContext<T> {
   static get<T>(contextName: string) {
     return getContext<QContext<T> | undefined>(contextName);
   }
-  
+
   /**
    * Prevents the propagation of the context further down
    * @param contextName - Name of the context to get
@@ -40,7 +40,7 @@ export default class QContext<T> {
   update(newVal: T) {
     this.#state = newVal;
   }
-  
+
   /**
    * Updates the property of id `key` with `newVal` in the context's inner value
    * @param newVal - New value to update the context's property with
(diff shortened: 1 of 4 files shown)


src/lib/classes/QContext.svelte.ts
```diff
@@ -1,26 +1,61 @@
 import { getContext, setContext } from "svelte";
 
-export class QContext<T> {
+/**
+ * This class allows to manipulate reactive contexts more easily.
+ * It avoid having to pass a Svelte store down the components but rather use runes to keep the context reactive.
```
therealcoder1337, line 5: `avoids`?

src/lib/classes/QContext.svelte.ts
```diff
@@ -1,26 +1,61 @@
...
   update(newVal: T) {
     this.#state = newVal;
   }
 
+  /**
+   * Updates the property of id `key` with `newVal` in the context's inner value
+   * @param newVal - New value to update the context's property with
+   */
+  updateEntry(key: keyof T, newVal: NonNullable<T>[keyof T]) {
+    if (!this.#state![key]) {
```
therealcoder1337, line 49: What if `this.#state.someKey === false` (or `null`, `0`, etc.)? Also: Do we really want to prevent setting values for new keys?

therealcoder1337 (changes requested):
Looks mostly clean, just two smaller things. Thanks!

src/lib/classes/QContext.svelte.ts
KilDesu: Oops, wrote too fast

src/lib/classes/QContext.svelte.ts
KilDesu: I should probably write `if(this.#state?.[key] === undefined)` then

therealcoder1337: Usually, you'd want to use [`Object.hasOwn()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn) then. You'd have to try if it works here.

src/lib/classes/QContext.svelte.ts
```diff
@@ -1,26 +1,61 @@
...
+  /**
+   * Updates the property of id `key` with `newVal` in the context's inner value
+   * @param newVal - New value to update the context's property with
+   */
+  updateEntry(key: keyof T, newVal: NonNullable<T>[keyof T]) {
+    if (!this.#state || typeof this.#state !== "object" || !Object.hasOwn(this.#state, key)) {
+      throw new Error(`${key.toString()} doesn't exist on ${this.#state}`);
+    }
+
+    this.#state![key] = newVal;
```
therealcoder1337, line 53: You could probably remove the `!` here, since you throw in the `if` above if `state` isn't set.

src/lib/classes/QContext.svelte.ts
```diff
@@ -1,26 +1,61 @@
...
     this.#state = newVal;
   }
 
+  /**
+   * Updates the property of id `key` with `newVal` in the context's inner value
+   * @param newVal - New value to update the context's property with
+   */
+  updateEntry(key: keyof T, newVal: NonNullable<T>[keyof T]) {
+    if (!this.#state || typeof this.#state !== "object" || !Object.hasOwn(this.#state, key)) {
+      throw new Error(`${key.toString()} doesn't exist on ${this.#state}`);
```
therealcoder1337, line 50: In case of missing/invalid `this.#state`, the error message is not correct.

therealcoder1337 (approved):

PR #78, opened 2024-05-23
Coder work

### LINKED ISSUE #76: isNumber refactor
There are two `isNumber`:
- `$utils/number.ts`
- `$utils/types.ts`

The one in `number.ts` seems to be the slightly improved variant.
So, suggestions:

- Remove the function in `types.ts` and replace occurrences
- Rename the other to `isNumeric` and adjust all calls
- Replace the parameter type `unknown` with `number | string` and use type guards around the function call where needed. Rationale: Something that is neither `number` nor `string` can never be numeric. We should also use `unknown` only where needed. If this last refactor causes too many complications, it can be omitted.

therealcoder1337:
## PR Type

- [x] Bugfix
- [x] Feature
- [ ] Documentation
- [x] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

## What's new?

> List the changes

- QSelect: fixes & new prop `displayValue`
- ESLint rule for import order (most notably, `type` imports will be last now)
- `isNumber`: remove duplicate code, rename to `isNumeric` and refactor
- Misc fixes/refactors

## Screenshots

N/A

Closes #76

### DIFF
--- a/src/lib/components/avatar/docs.ts
+++ b/src/lib/components/avatar/docs.ts
@@ -1,5 +1,5 @@
-import { QAvatarDocsProps } from "./docs.props";
 import type { QComponentDocs } from "$lib/utils";
+import { QAvatarDocsProps } from "./docs.props";
 
 export const QAvatarDocs: QComponentDocs = {
   name: "QAvatar",
--- a/src/lib/components/breadcrumbs/docs.ts
+++ b/src/lib/components/breadcrumbs/docs.ts
@@ -1,5 +1,5 @@
-import { QBreadcrumbsDocsProps } from "./docs.props";
 import type { QComponentDocs } from "$lib/utils";
+import { QBreadcrumbsDocsProps } from "./docs.props";
 
 export const QBreadcrumbsDocs: QComponentDocs = {
   name: "QBreadcrumbs",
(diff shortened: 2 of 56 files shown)


src/lib/components/avatar/docs.ts
```diff
@@ -1,5 +1,5 @@
-import { QAvatarDocsProps } from "./docs.props";
 import type { QComponentDocs } from "$lib/utils";
```
KilDesu, line 1: Didn't you say `import type` should be last?

src/lib/components/breadcrumbs/docs.ts
```diff
@@ -1,5 +1,5 @@
-import { QBreadcrumbsDocsProps } from "./docs.props";
 import type { QComponentDocs } from "$lib/utils";
```
KilDesu, line 1: Idem

src/lib/components/button/docs.ts
```diff
@@ -1,5 +1,5 @@
-import { QBtnDocsProps } from "./docs.props";
 import type { QComponentDocs } from "$lib/utils";
```
KilDesu, line 1: Idem, I'll stop mentioning it now

KilDesu (approved):
Good, thanks!

PR #93, opened 2024-09-03
feat: add a scroll observer utility class

KilDesu:
## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [ ] No

## What's new?

> List the changes

- Added an utility to track scroll

## Screenshots

N/A

N/A

### DIFF
--- a/src/lib/classes/QScrollObserver.svelte.ts
+++ b/src/lib/classes/QScrollObserver.svelte.ts
@@ -1,3 +1,7 @@
+import { elFromSelector } from "$utils";
+
+type Target = HTMLDivElement | typeof window | Document | null;
+
 type Direction = "up" | "right" | "down" | "left";
 
 type WindowScrollType = "scrollX" | "scrollY";
@@ -6,6 +10,12 @@ type DocumentScrollType = "scrollLeft" | "scrollTop";
 type ScrollType = {
   window: WindowScrollType;
   document: DocumentScrollType;
+};
+
+type Options = {
+  horizontal: boolean;
+  debounce: number;
+  ignore: (HTMLElement | null | string)[];
 };
 
 /**
@@ -22,13 +32,13 @@ type ScrollType = {
  * @param debounce - Time in milliseconds between each observation update
  */
 export default class QScrollObserver {
-  private scrollType: ScrollType = {
+  protected scrollType: ScrollType = {
     window: "scrollY",
     document: "scrollTop",
   };
-  private lastScroll: number;
-  private horizontal: boolean = false;
-  private clearTimer: (() => void) | null = null;
+  protected lastScroll: number;
+  protected horizontal: boolean = false;
+  protected clearTimer: (() => void) | null = null;
 
   direction = $state<Direction>("down");
   changedDirection = $state(false);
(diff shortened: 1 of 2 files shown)


src/lib/classes/QScrollObserver.svelte.ts
```diff
@@ -0,0 +1,109 @@
...
+ * const scroll = new QScrollObserver();
+ *
+ * $inspect(scroll.direction)
+ * ```
+ *
+ * @param horizontal - If set to `true`, observe horizontal scrolling rather than vertical
+ * @param debounce - Time in milliseconds between each observation update
+ */
+export default class QScrollObserver {
+  private scrollType: ScrollType = {
```
therealcoder1337, line 25: Can we, as a rule of thumb, please use `protected` instead of `private`? It will make inheriting much easier.

src/lib/classes/QScrollObserver.svelte.ts
```diff
@@ -0,0 +1,109 @@
...
+  private updateDirection(target: HTMLDivElement) {
+    this.clearTimer?.();
+
+    const newScrollPosition = this.getScrollPosition(target);
+
+    this.position = newScrollPosition;
+    this.delta = newScrollPosition - this.lastScroll;
+
+    const newDirection: Direction =
+      this.delta > 0 ? (this.horizontal ? "right" : "down") : this.horizontal ? "left" : "up";
```
therealcoder1337, line 98: Why is one in braces while the other isn't?

src/lib/classes/QScrollObserver.svelte.ts
```diff
@@ -0,0 +1,109 @@
...
+    this.inflectionPoint = 0;
+    this.direction = horizontal ? "right" : "down";
+
+    const handler = (e: Event) => {
+      const target = e.target as HTMLDivElement | null;
+      if (!target || target.parentElement?.classList.contains("q-layout")) {
+        return;
+      }
+
+      if (!debounce) {
```
therealcoder1337, line 57: I find this whole block a bit hard to grasp. Can we separate the debounce logic from this classes' logic?

therealcoder1337 (changes requested):
Looks clean and good overall, just a few minor things. 👍

src/lib/classes/QScrollObserver.svelte.ts
KilDesu: I don't see why this class should be inherited from but sure ig...

src/lib/classes/QScrollObserver.svelte.ts
KilDesu: No sure myself, seems like prettier just felt like it

src/lib/classes/QScrollObserver.svelte.ts
KilDesu: I can try but it'll be really annoying to do because this uses the class' functions. I guess I could pull this out with a `.bind(this, debounce, target)` but that'd be a hassle just for some debounce logic

KilDesu: I could separate it in a class function though

therealcoder1337 (approved):
Thanks!

PR #108, opened 2024-10-29
QBtn varant API change

KilDesu:
## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

## What's new?

> List the changes

- Added the possibility to choose a variant using boolean attributes
- To avoid variant collisions, I chose a hierarchy: `variant > filled > tonal > outlined > flat > elevated (default)`
- Fixed the problem with prettier eof breaking `bun lint` on Windows

## Screenshots

N/A

N/A

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -2,6 +2,7 @@
   "useTabs": false,
   "singleQuote": false,
   "trailingComma": "es5",
+  "endOfLine": "auto",
   "printWidth": 100,
   "plugins": ["prettier-plugin-svelte"],
   "overrides": [{ "files": "*.svelte", "options": { "parser": "svelte" } }]
--- a/.prettierrc
+++ b/.prettierrc
@@ -2,7 +2,7 @@
   "useTabs": false,
   "singleQuote": false,
   "trailingComma": "es5",
-  "endOfLine": "auto",
+  "endOfLine": "lf",
   "printWidth": 100,
   "plugins": ["prettier-plugin-svelte"],
   "overrides": [{ "files": "*.svelte", "options": { "parser": "svelte" } }]
--- a/src/lib/components/button/QBtn.svelte
+++ b/src/lib/components/button/QBtn.svelte
@@ -13,7 +13,11 @@
 
   let {
     disabled = false,
-    design = "elevated",
+    variant = "elevated",
+    filled = false,
+    tonal = false,
+    outlined = false,
+    flat = false,
     icon,
     label,
     loading = false,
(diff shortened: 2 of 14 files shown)


.prettierrc
```diff
@@ -2,6 +2,7 @@
   "useTabs": false,
   "singleQuote": false,
   "trailingComma": "es5",
+  "endOfLine": "auto",
```
therealcoder1337, line 5: Let's use `lf` please, which should evaluate to the same result but is more explicit.

src/lib/components/button/QBtn.svelte
```diff
@@ -39,8 +43,22 @@
 
   const src = $derived(extractImgSrc(icon));
 
+  const finalVariant = $derived<typeof variant>(
+    variant !== "elevated"
```
therealcoder1337, line 47: This is very hard to read. How about this (untested)?
```ts
const variants = {
  elevated,
  filled,
  tonal,
  outlined
}

return Object.keys(variants).find(key => variants[key])
```

src/lib/components/button/QBtn.svelte
```diff
@@ -39,8 +43,22 @@
...
+          : outlined
+            ? "outlined"
+            : flat
+              ? "flat"
+              : "elevated"
+  );
+
   const color = $derived(
-    `var(--${design === "filled" ? "on-primary" : design === "tonal" ? "on-secondary-container" : "primary"})`
+    `var(--${finalVariant === "filled" ? "on-primary" : finalVariant === "tonal" ? "on-secondary-container" : "primary"})`
```
therealcoder1337, line 61: Please always use braces for nested ternaries (or even better, avoid nested ternaries completely: https://eslint.org/docs/latest/rules/no-nested-ternary ). Maybe there's also an ESLint or Prettier rule for that.

src/lib/components/button/props.ts
```diff
@@ -13,10 +13,34 @@ export interface QBtnProps extends HTMLAttributes<HTMLButtonElement> {
...
   /**
-   * Choose the design for the button.
+   * Choose the variant for the button. If a variant is specified, it overwrites all other variants defined with boolean attributes.
    * @default "elevated"
    */
-  design?: QBtnDesignOptions;
+  variant?: QBtnDesignOptions;
+
+  /**
+   * Equivalent to `variant="filled"`. Overwritten by `variant` but overwrites `tonal`, `outlined` and `flat`.
```
therealcoder1337, line 22: It's great you've specified the order of importance.

therealcoder1337:
Thanks! A Few notes:
- I don't see any code snippets on the docs pages anymore, even after running `bun run docgen-snippets`. That might not be related to these changes.
- *After* this is approved, please tag the next version. To do that, first (and only once) add `tag-version-prefix=` to your `~/.npmrc` (to remove the default `v` prefix) and then execute `npm version 0.1.0-prealpha22` inside the project dir. Verify there is a new commit in `git log`. After pushing normally, you also need to push with the `--tags` option, e.g. `git push origin HEAD --tags`, since git tags are pushed separately.
- When the above is done and this branch is merged (not rebased), you can run the CI job to publish to npm.

therealcoder1337 (changes requested):
Please see comments.

.prettierrc
KilDesu: `lf` is the default for `endOfLine`, this exactly what creates the problem for Windows (I already tried to use `lf`, but it doesn't work)

therealcoder1337 (approved):

PR #111, opened 2024-11-25
Fix QTooltip to make it work

KilDesu:
## PR Type

- [x] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

## What's new?

> List the changes

- Fix QTooltip so it works with multiple positions
- Make tooltips attached to body and mount them only when needed
- Add a `data-quaff` attribute on the root element of all Quaff components so it's easier to find the right element to attach the tooltips to (will also help if we create a popover component)
- Fix QBtn children not being rendered if the `label` prop is defined (would not render tooltips if a label is defined)
- Remove unnecessary styles for QTooltip
- Possibility to hover the tooltip without making it disappear

**NB:** I had to create a `QTooltipBase` helper component to have something to imperatively `mount`. I also had to add a `q-tooltip__helper` div in the `QTooltip` component (the one exposed to users) to be able to find traverse parents. This implementation is very complex (and maybe ugly too), it might be possible to make it easier but at least it works now.

## Screenshots

N/A

N/A


src/lib/components/list/QItem.svelte
```diff
@@ -76,6 +77,7 @@
     {...Q.classes}
     tabindex={isClickable ? tabindex || 0 : undefined}
     aria-disabled={isActionable && disabled ? true : undefined}
+    data-quaff
```
therealcoder1337, line 80: Why a second time?

src/lib/components/button/QBtn.svelte
```diff
@@ -149,8 +150,8 @@
   <span class="q-btn__label">
     {#if label}
       {label}
-    {:else}
-      {@render children?.()}
     {/if}
+
```
therealcoder1337, line 154: This change will probably mean we can have both a `label` and contents set, which might not be desirable.

KilDesu: Yes but it's up to the user to not break things? Or maybe we can make `children` overwrite `label`? Might work too but is pretty bad DX when using tooltips

src/lib/components/list/QItem.svelte
KilDesu: What do you mean? Like "why in QItem if we use it on QList too"?

KilDesu:
We might be able to use Svelte's transitions to show and hide the tooltip instead of using CSS (because we mount and unmount it).
I'll try it today and see if it works.
I'll also remove code that might be rather useless and repetitive.

src/lib/components/button/QBtn.svelte
therealcoder1337: The safer solution would be better here imo, as it's not super clear that you are not supposed to use both the slot and the `label` attribute.

src/lib/components/list/QItem.svelte
therealcoder1337: Yeah

src/lib/components/tooltip/QTooltip.svelte
```diff
@@ -1,17 +1,183 @@
...
+
+  let timerShow = $state<ReturnType<typeof setTimeout> | null>(null);
+  let timerHide = $state<ReturnType<typeof setTimeout> | null>(null);
+
+  let mountedTooltip: ReturnType<typeof mountTooltip> | null = null;
+
+  const id = Date.now();
+
+  $effect(() => {
+    if (value) {
```
therealcoder1337, line 30: Could this be done without the `if` and a `ternary` instead or would it loose reactivity then?

src/lib/components/tooltip/QTooltip.svelte
```diff
@@ -1,17 +1,183 @@
...
+        if (parent.attributes.getNamedItem("data-quaff")) {
+          realTarget = parent;
+          break;
+        } else {
+          parent = parent.parentElement;
+        }
+      }
+    }
+
+    realTarget?.addEventListener("mouseenter", show);
```
therealcoder1337, line 56: We have this repeating three times, maybe it would be good to have a function to reduce code duplication. For example, the function could take args `el: HTMLElement` and `addOrRemove: 'add' | 'remove'` (didn't check types).

src/lib/components/tooltip/QTooltip.svelte
```diff
@@ -1,17 +1,183 @@
...
+    if (timerShow) {
+      clearTimeout(timerShow);
+    }
+
+    timerShow = setTimeout(() => {
+      timerShow = null;
+      mountedTooltip = mountTooltip();
+
+      tooltipEl = (document.getElementById(`qtooltip-${id}`) as HTMLDivElement) || undefined;
+      tooltipEl?.addEventListener("mouseenter", abortHide);
```
therealcoder1337, line 103: There's another one that could be replaced by mentioned function.

src/lib/components/tooltip/QTooltipBase.svelte
```diff
@@ -0,0 +1,70 @@
...
+    // 0.5rem offset so the tooltip is not stuck to the anchor
+    const defaultOffset = position.includes("-") ? 0 : 8;
+
+    const top = position.includes("top")
+      ? anchorRect.top - tooltipEl.offsetHeight + (offset.y || 0) - defaultOffset
+      : position.includes("bottom")
+        ? anchorRect.bottom + (offset.y || 0) + defaultOffset
+        : anchorRect.top + anchorRect.height / 2 - tooltipEl.offsetHeight / 2 + (offset.y || 0);
+
+    const left = position.includes("left")
```
therealcoder1337, line 41: It looks very similar to the above one. Can both be refactored into a function?

src/lib/components/tooltip/QTooltipBase.svelte
```diff
@@ -0,0 +1,70 @@
...
+      ? anchorRect.left - tooltipEl.offsetWidth + (offset.x || 0) - defaultOffset
+      : position.includes("right")
+        ? anchorRect.right + (offset.x || 0) + defaultOffset
+        : anchorRect.left + anchorRect.width / 2 - tooltipEl.offsetWidth / 2 + (offset.x || 0);
+
+    return { top, left };
+  });
+
+  const styles = $derived(
+    tooltipPosition && { top: `${tooltipPosition.top}px`, left: `${tooltipPosition.left}px` }
```
therealcoder1337, line 51: It's rather unusual to use `&&` to return a value. While you see `||` often, `&&` is not as easily understood for this by many. Maybe use a ternary instead (optional).

src/lib/components/list/QItem.svelte
KilDesu: Because users should be able to attach tooltips to any QItem imho, not only the full list

src/lib/components/tooltip/QTooltip.svelte
KilDesu: Wdym "and a ternary"? You mean writing `value ? untrack(show) : untrack(hide)`? Should be doable

src/lib/components/tooltip/QTooltip.svelte
KilDesu: Yeah, thought about that. Will do it

src/lib/components/tooltip/QTooltipBase.svelte
KilDesu: I guess so, yeah

src/lib/components/tooltip/QTooltipBase.svelte
KilDesu: Well, it's like writing `let activeClass = active && "my-class"`, it's exactly the same and we used it a lot so I think I'll keep it as is

src/lib/components/tooltip/QTooltip.svelte
KilDesu: Done, doesn't break anything

therealcoder1337 (approved):
GJ, thx!

PR #116, opened 2025-02-24
Snippets fix

### LINKED ISSUE #100: Find a good way to show custom types to user
As the title says, it would be great to find a good way to be able to show the users the definition of Quaff's custom types.
Today, we have a semi-working solution that stops working for nested types and when more or less complex union types are involved

KilDesu:
## PR Type

- [x] Bugfix
- [ ] Feature
- [x] Documentation
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

## What's new?

> List the changes

- Replaced slots by snippets in the API docs
- Made snippets docs generation automatic (as opposed to slots)
- Improved the type checker inside API
- Replaced drawer displaying types definition on click by a tooltip (on mouseenter)
- Fixed some components that didn't work
- Made all component docs work with the new docs gen api
- Import "shiki" when `code` changes inside `QCodeBlock` instead of onMount

## Screenshots

N/A

Closes #100

### DIFF
--- a/src/lib/components/list/props.ts
+++ b/src/lib/components/list/props.ts
@@ -1,6 +1,6 @@
 import type { RouterProps } from "$lib/utils/router";
 import type { Snippet } from "svelte";
-import type { QSeparatorProps } from "../separator/props";
+import type { QSeparatorHorizontalProps } from "../separator/props";
 import type { HTMLAnchorAttributes, HTMLAttributes } from "svelte/elements";
 
 export interface QListProps extends HTMLAttributes<HTMLElement> {
@@ -8,7 +8,7 @@ export interface QListProps extends HTMLAttributes<HTMLElement> {
   roundedBorders?: boolean;
   dense?: boolean;
   separator?: boolean;
-  separatorOptions?: Omit<QSeparatorProps, "vertical">;
+  separatorOptions?: Omit<QSeparatorHorizontalProps, "vertical">;
   padding?: boolean;
   tag?: string;
 }
--- a/src/lib/components/list/props.ts
+++ b/src/lib/components/list/props.ts
@@ -8,14 +8,7 @@ export interface QListProps extends HTMLAttributes<HTMLElement> {
   roundedBorders?: boolean;
   dense?: boolean;
   separator?: boolean;
-  separatorOptions?: {
-    spacing?: QSeparatorProps["spacing"];
-    inset?: QSeparatorProps["inset"];
-    color?: QSeparatorProps["color"];
-    size?: QSeparatorProps["size"];
-    text?: QSeparatorProps["text"];
-    textAlign?: QSeparatorProps["textAlign"];
-  };
+  separatorOptions?: Exclude<QSeparatorProps, "vertical">;
   padding?: boolean;
   tag?: string;
 }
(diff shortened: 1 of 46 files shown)


src/lib/components/list/props.ts
```diff
@@ -8,14 +8,7 @@ export interface QListProps extends HTMLAttributes<HTMLElement> {
...
   separator?: boolean;
-  separatorOptions?: {
-    spacing?: QSeparatorProps["spacing"];
-    inset?: QSeparatorProps["inset"];
-    color?: QSeparatorProps["color"];
-    size?: QSeparatorProps["size"];
-    text?: QSeparatorProps["text"];
-    textAlign?: QSeparatorProps["textAlign"];
-  };
+  separatorOptions?: Exclude<QSeparatorProps, "vertical">;
```
therealcoder1337, line 11: Not sure, but wouldn't this be `Omit` instead of `Exclude`?

KilDesu: True, you're right

src/lib/components/private/QApi.svelte
```diff
@@ -1,78 +1,154 @@
...
+            },
+          });
+        });
+      });
+    });
+  }
+
+  $effect(() => {
+    // Doesn't rerun if we don't use JSON.stringify
+    void JSON.stringify(api);
```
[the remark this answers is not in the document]
KilDesu: It's not obligatory but I found myself use it recently. I can remove it though.
I don't think there's any real difference to using it or not.

src/lib/components/private/QApi.svelte
```diff
@@ -1,78 +1,154 @@
...
+    ) => `<span${className(typeStyle, isClickable, typeName)}>${spanContent}</span>`;
+
+    let content = "<pre>";
+
+    if (prop.optional) {
+      content += inSpan("?", { typeStyle: true });
+    }
+
+    if (!Array.isArray(prop.type) || prop.type.length) {
+      content += inSpan(":\u00A0");
```
therealcoder1337, line 61: `\u00A0` is quite incomprehensible.

src/lib/components/private/QApi.svelte
```diff
@@ -1,78 +1,154 @@
...
+      content += inSpan(Array.isArray(prop.type) ? "(" : "", { typeStyle: true });
     }
 
-    if (!drawer[QDocument.name][doc.name]) {
-      setTimeout(() => {
-        drawerContent = content;
-        drawer[QDocument.name][doc.name] = true;
-      }, 100);
+    if (prop.isSnippet && Array.isArray(prop.type) && prop.type.length) {
+      content += inSpan("{\u00A0", { typeStyle: true });
```
therealcoder1337, line 69: Same here, can we have a variable that explains what `\u00A0` is?

src/lib/components/private/QApi.svelte
KilDesu: It's a space. I used it because a simple " " didn't work in the markup but now I moved the api creation inside JS, I can use a simple " " and remove those.

src/lib/components/private/QApi.svelte
KilDesu: Removed it

therealcoder1337 (approved):

PR #117, opened 2025-03-01
Docs: Add description and default where needed

KilDesu:
## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

## What's new?

> List the changes

- Added prop description and default for component props that didn't have them

## Screenshots

N/A

N/A

### DIFF
--- a/src/lib/components/separator/props.ts
+++ b/src/lib/components/separator/props.ts
@@ -1,22 +1,104 @@
 import type { HTMLAttributes } from "svelte/elements";
 
 export interface QSeparatorVerticalProps {
+  /**
+   * Spacing around the separator.
+   *
+   * @default "none"
+   */
   spacing?: Q.Size;
+
+  /**
+   * Adds horizontal padding to the separator container, adding space around the separator.
+   *
+   * @default false
+   */
   inset?: boolean;
+
+  /**
+   * Sets the separator orientation to vertical.
+   *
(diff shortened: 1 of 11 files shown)


src/lib/components/separator/props.ts
```diff
@@ -1,22 +1,104 @@
...
+
+  /**
+   * Adds vertical padding to the separator container, adding space around the separator.
+   *
+   * @default false
+   */
   inset?: boolean;
+
+  /**
+   * Sets the separator orientation to horizontal.
```
therealcoder1337, line 70: You mean `to vertical`?

therealcoder1337 (approved):
Just one comment to check, good otherwise. Thanks.

src/lib/components/separator/props.ts
KilDesu: Mmmh, I guess it could be seen like that, but I meant horizontal here because it's always `vertical: false` so it's horizontal. But I don't have strong feelings about it, do you think it's still better to go for "to vertical"?

therealcoder1337: Thought it was an error, let's keep it this way then.

PR #119, opened 2025-03-10
Refactor preprocessor to use new Svelte feature for dynamic classes

KilDesu:
## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

## What's new?

> List the changes

- Bump Svelte to latest version for the new feature to work
- Refactor the preprocessor to output `class={[...]}` rather than a combo of `class="..."` and `class:myClass="..."`
- Fix eslint watching vite config files by adding them to eslint ignore
- Remove the need to write `{...Q.classes}` for elements to be targeted
- Actually have type checking for the preprocessor code

## Screenshots

N/A

N/A


plugins/class-preprocessor/markup.js
```diff
@@ -20,7 +20,7 @@ export function prepareMarkup(fragment, component, uses, namespace) {
     } else if (node.type === "AwaitBlock") {
       /** @type {"pending" | "then" | "catch"} */
       let awaitKey;
-      for (awaitKey of ["pending", "then", "catch"]) {
+      for (awaitKey of /** @type {const} */ (["pending", "then", "catch"])) {
```
therealcoder1337, line 23: Could've used `as const` instead maybe?

plugins/class-preprocessor/markup.js
```diff
@@ -34,6 +34,10 @@ export function prepareMarkup(fragment, component, uses, namespace) {
       continue;
     }
 
+    if (!("fragment" in node)) {
```
therealcoder1337, line 37: Isn't this redundant with the following if?

therealcoder1337 (approved):
LGTM 👍

plugins/class-preprocessor/markup.js
KilDesu: Totally.

plugins/class-preprocessor/markup.js
KilDesu: Indeed.

PR #127, opened 2025-04-15
Finish components docs + some fixes

KilDesu:
## PR Type

- [x] Bugfix
- [ ] Feature
- [x] Documentation
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

## What's new?

> List the changes

- Finish writing components docs
- Add change event for QSwitch
- Remove unecessary prop in QTabs
- Fix styling problem on round QBtn that have a tooltip
- Add custom HEX color support for progress
- Fix some prop and style problems for progress
- Normalize progress values so both 0.5 and 50 (as a percentage) are accepted

## Screenshots

N/A

N/A

### DIFF
--- a/src/lib/components/button/QBtn.svelte
+++ b/src/lib/components/button/QBtn.svelte
@@ -1,6 +1,7 @@
 <svelte:options runes={true} />
 
 <script lang="ts">
+  import { onMount } from "svelte";
   import QCircularProgress from "$lib/components/progress/QCircularProgress.svelte";
   import { useSize } from "$lib/composables/useSize";
   import { ripple } from "$lib/helpers";
@@ -10,7 +11,6 @@
   import QIcon from "../icon/QIcon.svelte";
   import type { MaterialSymbol } from "material-symbols";
   import type { QBtnVariantOptions, QBtnProps } from "./props";
-  import { onMount } from "svelte";
 
   let {
     disabled = false,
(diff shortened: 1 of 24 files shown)


src/lib/components/button/QBtn.svelte
```diff
@@ -102,6 +104,15 @@
     stopIfDisabled(click);
   }
 
+  onMount(() => {
+    const { width, height } = qBtnLabel.getBoundingClientRect();
+
+    // This is required for buttons with no label and with a tooltip to be round
```
therealcoder1337, line 110: Can't we have a simpler solution for this?

src/lib/components/progress/QLinearProgress.svelte
```diff
@@ -23,6 +23,12 @@
   }: QLinearProgressProps = $props();
 
   const normalized = $derived(value > 1 ? value / 100 : value);
+  const normalizedBuffer = $derived(buffer && buffer > 1 ? buffer / 100 : buffer);
```
therealcoder1337, line 26:
```suggestion
  const normalizedBuffer = $derived(typeof buffer === 'number' && buffer > 1 ? (buffer / 100) : buffer);
```

src/lib/components/button/QBtn.svelte
KilDesu: If only... I wanted a CSS solution at first, but as text nodes can't be target by any selector, I didn't manage to find a better solution

therealcoder1337 (approved):

PR #134, opened 2025-05-08
refactor: use symbols instead of strings for context names

### LINKED ISSUE #113: Use `Symbol` each time we use context API
Right now, we use simple strings to name our contexts (maybe not all, but we sure do for some places)

This can cause problems as it exposes private API to the users:
Svelte's context is also given to slotted content so, if a user calls `getContext` with one of our contexts names, they can access the context and potentially mutate private properties that should not be manually mutated.

To ensure that users can't access those contexts, we should wrap our context names with a `Symbol()` and export them to get the context. This ensures that users won't be able to access them.

KilDesu:
As mentionned in #113, this is to avoid potential private API leakage.

## PR Type

- [x] Bugfix (potentially)
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

## What's new?

> List the changes

- Use globally exported symbols for context names instead of simple strings

## Screenshots

N/A

Closes #113

### DIFF
--- a/src/lib/classes/QContext.svelte.ts
+++ b/src/lib/classes/QContext.svelte.ts
@@ -9,28 +9,28 @@ import { getContext, setContext } from "svelte";
 export class QContext<T> {
   #state = $state<T>();
 
-  constructor(contextName: string, init: T) {
+  constructor(contextSymbol: symbol, init: T) {
     this.#state = init;
 
-    setContext(contextName, this);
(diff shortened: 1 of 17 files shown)


therealcoder1337 (approved):
LGTM, thanks. 👍

PR #172, opened 2025-08-16
feat: add a new context API

KilDesu:
## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [x] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

## What's new?

> List the changes

- Add a new preprocessor to facilitate working with Svelte contexts. It automatically adds getters and setters for variables passed in contexts, maintaining proper reactivity
- Update the code to use the new preprocessor and context API
- Wrap code in foldable regions to navigate the code more easily

## Screenshots

N/A

N/A


plugins/contextPreprocessor.js
therealcoder1337, line 1: The code in this file is not very easy to understand, it might be worth to overhaul it (use smaller functions that that ideally follow the single responsibility principle).

src/lib/components/checkbox/QCheckbox.svelte
```diff
@@ -1,7 +1,9 @@
 <script lang="ts">
   import type { QCheckboxProps } from "./props";
 
+  // #region:    --- Props
   let { value = $bindable(), label = "", disable = false, ...props }: QCheckboxProps = $props();
+  // #endregion: --- Props
```
therealcoder1337, line 6: Is it intended for some components to just have a single region?

src/lib/components/chip/QChip.svelte
```diff
@@ -7,23 +7,39 @@
...
+  // #region:    --- Non-reactive variables
   let qChip: HTMLDivElement;
+  // #endregion: --- Non-reactive variables
+
+  // #region:    --- Derived values
+  const trailing = $derived(
+    (kind === "assist" || kind === "suggestion") && trailingIcon ? undefined : trailingIcon
+  );
+
+  const tabindex = disabled ? -1 : props.tabindex || 0;
```
therealcoder1337, line 36:
```suggestion
  const tabindex = disabled ? -1 : (props.tabindex || 0);
```

src/lib/components/layout/QLayout.svelte
```diff
@@ -33,87 +43,130 @@
...
-
-  const topOffset = $derived(!header || headerCtx.value.collapsed ? 0 : headerCtx.value.height);
-  const bottomOffset = $derived(!footer || footerCtx.value.collapsed ? 0 : footerCtx.value.height);
-  const leftOffset = $derived(handleDrawerCtx(leftRailbarCtx) + handleDrawerCtx(leftDrawerCtx));
-  const rightOffset = $derived(handleDrawerCtx(rightRailbarCtx) + handleDrawerCtx(rightDrawerCtx));
+  // #endregion: --- Reactive variables
+
+  // #region:    --- Derived values
+  const topOffset = $derived(!header || headerInfo.collapsed ? 0 : headerInfo.height);
+  const bottomOffset = $derived(!footer || footerInfo.collapsed ? 0 : footerInfo.height);
```
therealcoder1337, line 86: Readability could be improved here, e.g. by adding braces.

src/lib/components/layout/QLayout.svelte
```diff
@@ -33,87 +43,130 @@
...
         (layoutEl && !layoutEl.querySelector(":scope > .q-railbar--right"))) &&
       (!drawerLeft ||
-        leftDrawerCtx.value.ready ||
+        leftDrawerInfo.ready ||
         (layoutEl && !layoutEl.querySelector(":scope > .q-drawer--left"))) &&
       (!drawerRight ||
-        rightDrawerCtx.value.ready ||
+        rightDrawerInfo.ready ||
         (layoutEl && !layoutEl.querySelector(":scope > .q-drawer--right")))
   );
```
therealcoder1337, line 111: This could also be refactored to be a bit more readable.

src/lib/components/layout/QLayout.svelte
```diff
@@ -33,87 +43,130 @@
...
+    height: headerInfo.height,
+    collapsed: headerInfo.collapsed,
+    ready: headerInfo.ready,
+  });
+  footerCtx.set({
+    view,
+    height: footerInfo.height,
+    collapsed: footerInfo.collapsed,
+    ready: footerInfo.ready,
+  });
```
therealcoder1337, line 126: There could be a function (2 args) that does this and takes `headerCtx`/`footerCtx` and `headerInfo`/`footerInfo` as arguments.

src/lib/components/layout/QLayout.svelte
```diff
@@ -33,87 +43,130 @@
...
+    width: leftDrawerInfo.width,
+    takesSpace: leftDrawerInfo.takesSpace,
+    ready: leftDrawerInfo.ready,
+  });
+  rightDrawerCtx.set({
+    view,
+    width: rightDrawerInfo.width,
+    takesSpace: rightDrawerInfo.takesSpace,
+    ready: rightDrawerInfo.ready,
+  });
```
therealcoder1337, line 152: These could also be refactored to use a function similar to above, saves a lot of repetition.

src/lib/components/private/QDocsSection.svelte
```diff
@@ -10,16 +10,22 @@
...
+  // #region:    --- Reactive variables
+  let dialog = $state(false);
 
-  const code = $derived(snippets && snippets()[title]);
+  const ctx = docsCtx.get();
+  // #endregion: --- Reactive variables
 
-  let dialog = $state(false);
+  // #region:    --- Derived values
+  const code = $derived(ctx?.snippets && ctx.snippets[title]);
```
therealcoder1337, line 24:
```suggestion
  const code = $derived(ctx?.snippets?.[title]);
```
Wouldn't this style work? I feel we have a lot of `&&` where it might not be needed, and it's also not super nice for readability.

src/lib/components/progress/QCircularProgress.svelte
```diff
@@ -33,7 +29,16 @@
     children = fallback,
     ...props
   }: QCircularProgressProps = $props();
+  // #endregion: --- Props
+
+  // #region:    --- Non-reactive variables
+  const radius = 50,
+    diameter = 2 * radius,
+    circumference = diameter * Math.PI,
+    strokeDashArray = Math.round(circumference * 1000) / 1000;
```
therealcoder1337, line 38:
```suggestion
  const radius = 50;
  const diameter = 2 * radius;
  const circumference = diameter * Math.PI;
  const strokeDashArray = Math.round(circumference * 1000) / 1000;
```
Both styles are valid, this is just a suggestion to stay consistent. Or do we use comma initialization often?

src/lib/components/progress/QLinearProgress.svelte
```diff
@@ -37,6 +36,13 @@
   );
   const trackTransform = $derived(width(normalizedBuffer ?? 1, reverse));
   const indicatorTransform = $derived(width(+indeterminate || normalized, reverse));
+  // #endregion: --- Derived values
+
+  // #region:    --- Functions
+  function width(val: number, reverse: boolean) {
+    return reverse ? `translateX(100%) scale3d(-${val}, 1, 1)` : `scale3d(${val}, 1, 1)`;
```
therealcoder1337, line 43: Unnecessary duplication of the `scale3d` part.

src/lib/components/checkbox/QCheckbox.svelte
KilDesu: It's not really required, but I did it just to be consistent. I can of course remove one-line regions if you think it's better

src/lib/components/chip/QChip.svelte
KilDesu: Prettier won't let me do so

src/lib/components/layout/QLayout.svelte
KilDesu: idem, prettier would add the braces if it's needed, it will remove them otherwise so I can't really "manually" add them without adding `prettier-ignore`

src/lib/components/layout/QLayout.svelte
KilDesu: That will not work because this piece of code (`myContext.set({ ... })`) is preprocessed and the variable declaration is used in the preprocessor checks.
With normal Svelte, without preprocessing, the code would actually look like this:
```js
headerCtx.set({
    get view() { return view },
    get height() { return headerInfo.height },
    set height(newValue) { headerInfo.height = newValue },
    get collapsed() { return headerInfo.collapsed },
    set collapsed(newValue) { headerInfo.collapsed = newValue },
    get ready() { return headerInfo.ready },
    set ready(newValue) { headerInfo.ready = newValue }
  });

// etc...
```
So I think it's ok to repeat this.
What I could do though, is to use an `info` property instead of three `height`, `collapsed` and `ready` properties

src/lib/components/layout/QLayout.svelte
KilDesu: idem, see my last response

src/lib/components/private/ContextReseter.svelte
```diff
@@ -11,7 +10,7 @@
   } = $props();
 
   const keysArr = Array.isArray(keys) ? keys : [keys];
-  keysArr.forEach((key) => QContext.reset(key));
+  keysArr.forEach((key) => setContext(key, undefined));
```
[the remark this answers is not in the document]
KilDesu: It's the right way to reset contexts because the initial value for contexts is indeed `undefined`. We're basically "removing" the contexts here so `null` would not make sense imo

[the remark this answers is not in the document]
KilDesu: Also, if you check `hasContext` with a `null` context, I think it would return `true` even though we would expect `false`

src/lib/components/private/QDocsSection.svelte
KilDesu: Yes it can (and will) be changed

src/lib/components/progress/QCircularProgress.svelte
KilDesu: I don't have a strong opinion on this, I can go with your code

src/lib/components/progress/QLinearProgress.svelte
KilDesu: There's a `-` (minus sign) on the second one, it's not 100% the same. I can avoid repeating `scale3d` if you wish me to, but it will add more code than there is now

src/lib/components/checkbox/QCheckbox.svelte
therealcoder1337: No, the question was: Should there be more than just one region?

plugins/contextPreprocessor.js
KilDesu: Done

src/lib/components/progress/QLinearProgress.svelte
therealcoder1337: Right, sorry.

plugins/contextPreprocessor.js
therealcoder1337, line 1: This should be TypeScript.

tsconfig.json
```diff
@@ -11,10 +11,10 @@
     "skipLibCheck": true,
     "sourceMap": true,
     "strict": true,
-    "moduleResolution": "node",
+    "moduleResolution": "bundler",
```
therealcoder1337, line 14: Is this recommended by SvelteKit?

therealcoder1337 (approved):

PR #184, opened 2026-05-20
fix(QScrollObserver,QFooter,QHeader): bugfixes & enhancements

therealcoder1337:
## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

## What's new?

> List the changes

- Fixed QScrollObserver listener leak (capture) + window target bug
- Refactored header/footer with reveal scroll helper (no more $derived leaks)

## Screenshots

N/A

N/A

### DIFF
--- a/src/lib/components/header/QHeader.svelte
+++ b/src/lib/components/header/QHeader.svelte
@@ -1,6 +1,6 @@
 <script lang="ts">
   import { onMount } from "svelte";
-  import { QScrollObserver } from "$lib";
+  import { useRevealScrollObserver } from "$composables";
   import { headerCtx } from "../layout/QLayout.svelte";
   import type { QHeaderProps } from "./props";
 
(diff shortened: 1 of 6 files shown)


KilDesu:
The `reveal` property of both QHeader and QFooter are broken due to an infinite effect loop. Maybe because `scroll` is both read and set in the `$effect` inside the `useRevealScrollObserver` composable?

KilDesu (changes requested):
Gotta fix the infinite effect loop

src/lib/components/header/QHeader.svelte
```diff
@@ -28,14 +28,19 @@
   // #endregion: --- Props
 
   // #region:    --- Derived values
-  const scroll = $derived(
-    reveal ? new QScrollObserver(`.q-header--${uid} ~ .q-layout__content`) : undefined
+  const revealObserver = useRevealScrollObserver(
+    () => reveal,
+    () => uid,
+    "header"
   );
```
KilDesu: I don't think `uid` needs to be in its own scope (it's not reactive), also I'd prefer this order:
```ts
  const revealObserver = useRevealScrollObserver(
    "header",
    uid,
    () => reveal
  );
```

src/lib/composables/useRevealScrollObserver.svelte.ts
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Manages a {@link QScrollObserver} for QHeader / QFooter `reveal` mode.
+ * Observes the sibling `.q-layout__content` scroll area and tears down on disable or unmount.
+ */
+export function useRevealScrollObserver(
+  reveal: () => boolean,
+  uid: () => string,
+  bar: RevealScrollBar
+) {
```
KilDesu: See my previous comment on the scope of `uid` and the order of the parameters (which make more sense to me, but it's not super important).

src/lib/composables/useRevealScrollObserver.svelte.ts
```diff
@@ -0,0 +1,47 @@
...
+  let scroll = $state<QScrollObserver | undefined>();
+  let currentScroll: QScrollObserver | undefined;
```
KilDesu: I'm not sure why `currentScroll` is needed and why `scroll` is a `$state` (especially when it's return with a getter).
Could you enlighten me?

KilDesu (approved):

therealcoder1337:
Thanks for the review!

PR #190, opened 2026-06-05
feat(QSelect): add useInput and filter features, refactors

therealcoder1337:
## PR Type

- [x] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

## What's new?

> List the changes

- `QSelect`: add filter features & begin refactoring parts of the code to be in separate files, to not bloat the component so much
- `QSelect`: fix some docs display issues
- `QSelect`/`QInput` fix small style issue when hovering the `filled` variants

## Screenshots

N/A

N/A

### DIFF
--- a/src/lib/components/select/QSelect.svelte
+++ b/src/lib/components/select/QSelect.svelte
@@ -1,7 +1,16 @@
 <script lang="ts">
-  import { tick } from "svelte";
+  import { onDestroy, tick } from "svelte";
   import { QIcon, QItem, QItemSection, QList, QMenu } from "$lib";
   import { isActivationKey, type QEvent } from "$utils";
+  import { createFilterRunner, getFilteredOptions, getInputValue } from "./filter";
+  import {
+    doValuesMatch,
+    getDisplayValue,
+    getInitialOptionIndex,
+    getNextValue,
+    getOptionLabel,
+    normalizeOptionIndex,
+  } from "./option";
   import type { QSelectOption, QSelectProps } from "./props";
 
   type QSelectEvent<T> = QEvent<T, HTMLDivElement>;
@@ -21,10 +30,15 @@
     rounded = false,
     displayValue,
     emitValue = false,
-    before = undefined,
-    prepend = undefined,
-    append = undefined,
-    after = undefined,
+    useInput = false,
+    filterable = false,
+    inputDebounce = 300,
+    noOptionText = "No options",
+    onFilter,
+    before,
+    prepend,
+    append,
+    after,
     value = $bindable(),
     ...props
   }: QSelectProps = $props();
@@ -42,25 +56,29 @@
   let isMenuOpen = $state(false);
   let focusedOptionIndex = $state(-1);
   let snippetPrependWidth = $state(0);
+  let searchValue = $state("");
+  let isSearching = $state(false);
   // #endregion: --- Reactive variables
 
-  // #region:    --- Derived values
-  const currentDisplayValue = $derived.by(() => {
-    if (displayValue !== undefined) {
-      return displayValue;
-    }
-
-    if (!multiple || !Array.isArray(value)) {
-      return getOptionLabel(value as QSelectOption);
-    }
-
-    return value.map(getOptionLabel).join(", ");
+  const filterRunner = createFilterRunner({
+    onAbort: resetFocusedOption,
+    onDone: focusFirstOption,
   });
+
+  // #region:    --- Derived values
+  const currentDisplayValue = $derived(getDisplayValue(value, options, multiple, displayValue));
 
   const hasDisplayValue = $derived(currentDisplayValue !== "" && currentDisplayValue !== undefined);
   const isActive = $derived(hasDisplayValue || isFocused || isMenuOpen);
 
-  const isOptionSelectedByIndex: boolean[] = $derived(options.map((option) => isSelected(option)));
+  const visibleOptions = $derived(
+    getFilteredOptions(options, searchValue, useInput, filterable, !!onFilter)
+  );
+  const inputValue = $derived(
+    getInputValue(currentDisplayValue, searchValue, useInput, isSearching)
+  );
+
+  const isOptionSelectedByIndex = $derived(visibleOptions.map(isSelected));
   const activeOptionId = $derived(
     isMenuOpen && focusedOptionIndex >= 0 ? getOptionId(focusedOptionIndex) : undefined
   );
@@ -76,23 +94,32 @@
   $effect(() => {
     if (!isMenuOpen) {
       focusedOptionIndex = -1;
+      resetSearch();
+    }
+  });
+
+  $effect(() => {
+    if (focusedOptionIndex >= visibleOptions.length) {
+      focusedOptionIndex = visibleOptions.length ? visibleOptions.length - 1 : -1;
     }
   });
   // #endregion: --- Effects
 
+  onDestroy(filterRunner.clear);
+
   // #region:    --- Functions
-  function handleMousedown(e: QSelectEvent<MouseEvent>) {
+  function handleMousedown() {
     if (disable) {
       return;
     }
 
-    if (isMenuOpen) {
+    if (useInput) {
+      void showMenu();
+    } else if (isMenuOpen) {
       hideMenu();
     } else {
       void showMenu();
     }
-
-    props.onmousedown?.(e);
   }
 
   function handleFocus(e: QSelectEvent<FocusEvent>) {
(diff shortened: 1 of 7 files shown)


src/lib/components/select/filter.ts
```diff
@@ -0,0 +1,109 @@
...
+        return;
+      }
+      aborted = true;
+      if (currentRun === run) {
+        onAbort();
+      }
+    };
+
+    try {
+      void Promise.resolve(onFilter(search, update, abort)).catch(() => abort());
```
KilDesu:
```suggestion
      void Promise.resolve(onFilter(search, update, abort)).catch(abort);
```

src/lib/components/select/option.ts
```diff
@@ -0,0 +1,95 @@
...
+  a: QSelectOption | null | undefined,
+  b: QSelectOption | null | undefined
+) {
+  return getOptionValue(a) === getOptionValue(b);
+}
+
+export function getOptionLabel(option: QSelectOption | null | undefined, options: QSelectOption[]) {
+  if (option === undefined || option === null || option === "") {
+    return "";
+  }
```
KilDesu, line 21:
```suggestion
  if (!option) {
    return "";
  }
```

src/lib/components/select/option.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  if (typeof option !== "string" && typeof option !== "number") {
+    return option.label;
+  }
+
+  return options.includes(option)
+    ? option
+    : (options.find((opt) => doValuesMatch(opt, option)) as { label: string | number })?.label ||
```
KilDesu: Wtf is this type casting?

src/lib/components/select/option.ts
```diff
@@ -0,0 +1,95 @@
...
+
+export function getDisplayValue(
+  value: QSelectValue | undefined,
+  options: QSelectOption[],
+  multiple: boolean,
+  displayValue?: string
+) {
+  if (displayValue !== undefined) {
+    return displayValue;
+  }
```
KilDesu, line 43:
```suggestion
  if (displayValue) {
    return displayValue;
  }
```
Do we really want to show `displayValue` if it's `""`?

src/lib/components/select/option.ts
```diff
@@ -0,0 +1,95 @@
...
+export function getInitialOptionIndex(
+  options: QSelectOption[],
+  isSelected: (option: QSelectOption) => boolean,
+  direction = 1
+) {
+  if (!options.length) {
+    return -1;
+  }
+
+  const selectedIndex = options.findIndex((option) => isSelected(option));
```
KilDesu:
```suggestion
  const selectedIndex = options.findIndex(isSelected);
```

src/lib/components/select/QSelect.svelte
```diff
@@ -21,6 +31,11 @@
...
     displayValue,
     emitValue = false,
+    useInput = false,
+    filterable = false,
+    inputDebounce = 300,
+    noOptionText = "No options",
+    onFilter,
     before = undefined,
     prepend = undefined,
     append = undefined,
```
KilDesu:
```suggestion
    before,
    prepend,
    append,
```

src/lib/components/select/QSelect.svelte
```diff
@@ -42,25 +57,31 @@
...
-  const isOptionSelectedByIndex: boolean[] = $derived(options.map((option) => isSelected(option)));
+  const visibleOptions = $derived(
+    getFilteredOptions(options, searchValue, useInput, filterable, !!onFilter)
+  );
+  const inputValue = $derived(
+    getInputValue(currentDisplayValue, searchValue, useInput, isSearching)
+  );
+
+  const isOptionSelectedByIndex: boolean[] = $derived(
+    visibleOptions.map((option) => isSelected(option))
```
KilDesu:
```suggestion
    visibleOptions.map(isSelected)
```

src/lib/components/select/QSelect.svelte
```diff
@@ -42,25 +57,31 @@
...
 
-  const isOptionSelectedByIndex: boolean[] = $derived(options.map((option) => isSelected(option)));
+  const visibleOptions = $derived(
+    getFilteredOptions(options, searchValue, useInput, filterable, !!onFilter)
+  );
+  const inputValue = $derived(
+    getInputValue(currentDisplayValue, searchValue, useInput, isSearching)
+  );
+
+  const isOptionSelectedByIndex: boolean[] = $derived(
```
KilDesu:
```suggestion
  const isOptionSelectedByIndex = $derived(
```
Do we really need an explicit type here?

src/lib/components/select/QSelect.svelte
```diff
@@ -76,23 +97,34 @@
...
+  });
+
+  $effect(() => {
+    if (focusedOptionIndex >= visibleOptions.length) {
+      focusedOptionIndex = visibleOptions.length ? visibleOptions.length - 1 : -1;
     }
   });
   // #endregion: --- Effects
 
+  onDestroy(() => filterRunner.clear());
```
KilDesu:
```suggestion
  onDestroy(filterRunner.clear);
```
Would it break things?

src/lib/components/select/QSelect.svelte
```diff
@@ -76,23 +97,34 @@
...
+    if (useInput) {
+      void showMenu();
+    } else if (isMenuOpen) {
       hideMenu();
     } else {
       void showMenu();
     }
 
-    props.onmousedown?.(e);
+    props.onmousedown?.(e as QSelectEvent<MouseEvent>);
```
KilDesu: Why was type casting not needed before but needed now? Is it that important to define `e` to be `QSelectMouseEvent`?

src/lib/components/select/option.ts
therealcoder1337: Yeah why not? People might want to display an empty string sometimes.

src/lib/components/select/option.ts
therealcoder1337: Apparently this would break `0`, so I'm keeping it as is.

src/lib/components/select/filter.ts
therealcoder1337: Done!

src/lib/components/select/option.ts
therealcoder1337: Refactored!

src/lib/components/select/option.ts
therealcoder1337: Refactored.

src/lib/components/select/QSelect.svelte
therealcoder1337: Done!

src/lib/components/select/QSelect.svelte
therealcoder1337: Refactored!

src/lib/components/select/QSelect.svelte
therealcoder1337: Refactored!

src/lib/components/select/QSelect.svelte
therealcoder1337: Refactored!

src/lib/components/select/QSelect.svelte
therealcoder1337: Code gone via refactor

src/lib/components/select/option.ts
KilDesu: > Apparently this would break `0`, so I'm keeping it as is.

Then what about
```suggestion
  if (option !== 0 && !option) {
    return "";
  }
```

src/lib/components/select/QSelect.svelte
```diff
@@ -76,23 +94,32 @@
...
+    if (useInput) {
+      void showMenu();
+    } else if (isMenuOpen) {
       hideMenu();
     } else {
       void showMenu();
     }
-
-    props.onmousedown?.(e);
   }
```
KilDesu, line 123: Somehow the event is still forwarded if a user uses `<QSelect onmousedown={(e) => console.log(e)} />`, it will work.
I don't know why, but seems like removing the `props.onevent` will not break things. Might be worth looking into refactoring to remove all our event forwards.

KilDesu (approved):

src/lib/components/select/option.ts
therealcoder1337: We can consider it another time, as I'm not sure if there's more side effects we didn't think about with the new solution.

PR #199, opened 2026-06-10
feat: improve page style, fix bugs, add docs 

therealcoder1337:
## PR Type

- [x] Bugfix
- [x] Feature
- [x] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

## What's new?

> List the changes

- improve page style
- add missing `q-col-gutter-*` classes
- add layout documentation, improve docs layout
- remove toolbar gap to prevent huge spacing
- remove stray char from `QHeader` docs

## Screenshots

N/A

N/A

### DIFF
--- a/src/lib/components/private/QApi.svelte
+++ b/src/lib/components/private/QApi.svelte
@@ -223,10 +223,7 @@
             <QItemSection type="content">
               {#snippet headline()}
                 <div class="q-api__doc-heading q-my-sm">
-                  <span
-                    class="q-px-sm q-py-xs q-mr-xs"
-                    style="background-color: var(--surface-container); border-radius: 0.5rem"
-                  >
+                  <span class="q-docs-code q-mr-xs">
                     <b>{doc.name}</b>
                   </span>
                   {#if isProp(doc, index)}
--- a/src/lib/components/private/QDocs.svelte
+++ b/src/lib/components/private/QDocs.svelte
@@ -51,7 +51,7 @@
 <div
   class="q-docs"
   style="--q-hue-rotate: {hueRotate}deg; --q-brightness: {brightness}; max-width: {Quaff.breakpoints
-    .lg}px; margin: auto;"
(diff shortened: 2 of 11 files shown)


src/lib/components/private/QApi.svelte
```diff
@@ -22,10 +22,18 @@
   // #endregion: --- Props
 
   // #region:    --- Reactive variables
-  let api: (keyof QComponentDocs["docs"])[] = $state(componentDocs.map(() => "props"));
+  let api: (keyof QComponentDocs["docs"])[] = $state([]);
   // #endregion: --- Reactive variables
 
   // #region:    --- Effects
+  $effect(() => {
+    if (api.length === componentDocs.length) {
```
therealcoder1337: Fixed a warning, hence the many changes in this file:
```
[vite-plugin-svelte] src/lib/components/private/QApi.svelte:30:4 This reference only captures the initial value of componentDocs. Did you mean to reference it inside a derived instead?
https://svelte.dev/e/state_referenced_locally
```

KilDesu: Although a prop, `componentDocs` is not reactive because it's data imported from the `docs.ts` files inside the components' folder, so it won't change during runtime.
To avoid this warning, we probably want to use context API (`createContext` for this component to access the data (I can't really propose changes because it needs changes to other files).

KilDesu (commented):
Can you take a look at my comment and see if you can change it? If not, I can try to create a PR myself when I get the time. I don't think we can merge this PR as it is, it makes changes we'd need to revert afterwards.

src/lib/components/private/QApi.svelte
therealcoder1337: I've removed this code for now, so you can add your fix later.

KilDesu (approved):
Thanks! Now we will want to actually use `q-page` in our docs

PR #201, opened 2026-06-10
fix: use "disabled" attr name, not "disable"

therealcoder1337:
## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes
- [ ] No

## What's new?

> List the changes

We've inconsistently mixed the attribute name `disabled` with `disable` to maintain stronger API compatibility with Quasar. It seems however there is no good reason to use `disable` and we'd like Quaff to follow the more generic standard attribute name `disabled` instead.

## Screenshots

N/A

N/A

### DIFF
--- a/src/lib/components/checkbox/QCheckbox.svelte
+++ b/src/lib/components/checkbox/QCheckbox.svelte
@@ -2,18 +2,18 @@
   import type { QCheckboxProps } from "./props";
 
   // #region:    --- Props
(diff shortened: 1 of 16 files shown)


KilDesu (approved):
Thanks!

PR #207, opened 2026-06-13
feat: change Quaff.init() to adjust display mode body class

therealcoder1337:
## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [ ] No

## What's new?

> List the changes

- change `Quaff.init()` to adjust display mode body class

Background: We'd like to persist dark mode settings in Quaff created projects.

## Screenshots

N/A

Depends on https://github.com/quaffui/create-quaff/pull/20.

### DIFF
--- a/src/app.html
+++ b/src/app.html
@@ -8,21 +8,19 @@
   </head>
   <body data-sveltekit-preload-data="hover" class="">
     <script type="module">
-      // Set dark mode based on system preference
-      const cachedScheme = localStorage.getItem("current_mode");
+      // prevents a white flash on load for dark mode users
+      const savedDisplayMode = localStorage.getItem("displayMode");
+      let currentDisplayMode;
 
-      if (cachedScheme) {
-        document.body.classList.add(`body--${cachedScheme}`);
+      if (savedDisplayMode === "dark" || savedDisplayMode === "light") {
+        currentDisplayMode = savedDisplayMode;
+      } else if (window.matchMedia("(prefers-color-scheme: dark)").matches) {
+        currentDisplayMode = "dark";
       } else {
-        const prefersDarkScheme = window.matchMedia("(prefers-color-scheme: dark)");
-
-        if (prefersDarkScheme.matches) {
-          document.body.classList.add("body--dark");
-          localStorage.setItem("current_mode", "dark");
-        } else {
-          localStorage.setItem("current_mode", "light");
-        }
+        currentDisplayMode = "light";
       }
+
+      document.body.classList.add(`body--${currentDisplayMode}`);
     </script>
 
     <div>%sveltekit.body%</div>
--- a/src/lib/classes/Quaff.svelte.ts
+++ b/src/lib/classes/Quaff.svelte.ts
@@ -3,7 +3,9 @@ import { innerWidth } from "svelte/reactivity/window";
 import { version } from "$helpers";
 import { page } from "$app/state";
 
-type Mode = "light" | "dark";
+type DisplayMode = "light" | "dark";
+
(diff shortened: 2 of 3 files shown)


src/lib/classes/Quaff.svelte.ts
```diff
@@ -62,33 +64,39 @@ class Quaff {
...
-    const body = document.querySelector("body");
-    body?.classList.replace(`body--${mode.from}`, `body--${mode.to}`);
-    localStorage.setItem("current_mode", mode.to);
+    if (typeof document !== "undefined") {
+      document.body.classList.remove("body--light", "body--dark");
+      document.body.classList.add(`body--${displayMode}`);
+    }
+
+    if (persist && typeof localStorage !== "undefined") {
+      localStorage.setItem(DISPLAY_MODE_STORAGE_KEY, displayMode);
```
KilDesu, line 90:
```suggestion
    if (persist) {
      localStorage?.setItem(DISPLAY_MODE_STORAGE_KEY, displayMode);
```

src/lib/classes/Quaff.svelte.ts
KilDesu, line 112:
```suggestion
public get darkMode() {
  return {
    isActive: () => this.dark,
    toggle: () => this.toggleDarkMode(),
    set: (newVal: boolean) => this.setDarkMode(newVal),
  };
}
```
Do you think it would be better to have this and use `isActive` as a function like `Quaff.darkMode.isActive()`? It avoids disabling eslint and `isActive` seems more semantically correct as a function imho.
Do what you think is better, no need to change it imperatively.

src/app.html
```diff
@@ -8,21 +8,19 @@
...
-        const prefersDarkScheme = window.matchMedia("(prefers-color-scheme: dark)");
-
-        if (prefersDarkScheme.matches) {
-          document.body.classList.add("body--dark");
-          localStorage.setItem("current_mode", "dark");
-        } else {
-          localStorage.setItem("current_mode", "light");
-        }
+        currentDisplayMode = "light";
       }
```
KilDesu, line 21: Don't we persist the display mode in local storage if it's not already there? Why is this?

KilDesu (commented):
Thanks! Some comments I'd like you to take a look at before I approve though.

src/lib/classes/Quaff.svelte.ts
therealcoder1337: Tbh I wouldn't do this as it may risk breaking SSR again.

src/app.html
therealcoder1337: There's a comment at the top. This code is to prevent the "white flash" you can get an initial loading if you have dark mode set.

src/lib/classes/Quaff.svelte.ts
therealcoder1337: Let's keep it like it is for now, for API compatibility. Although your point is valid.

src/app.html
KilDesu: I know for the white flash, I originally wrote the code in the html directly exactly to avoid this problem. That's besides the point though, I asked why don't we `localStorage.setItem` in this code to persist the `currentDisplayMode`? It would avoid another `window.matchMedia` check when `Quaff.init` fires.

therealcoder1337: Imo this is the responsibility of the `Quaff.init()` and we shouldn't scatter it by having two code parts that handle persistence. It's annoying enough that we need this flash fix, maybe we can find better ways in the future.

KilDesu (approved):

PR #208, opened 2026-06-14
refactor: upgrade ripple helper to use attachments and improve its co…

KilDesu:
…lor API

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [ ] No

## What's new?

> List the changes

- Migrate ripple helper to Svelte 5's attachment API
- Improve ripple color option for better UX (when we open its API to the lib)
- Refactor components to use the new ripple API

## Screenshots

N/A

N/A

### DIFF
--- a/src/lib/helpers/ripple.ts
+++ b/src/lib/helpers/ripple.ts
@@ -1,70 +1,95 @@
+import { Attachment } from "svelte/attachments";
 import { on } from "svelte/events";
+import { isActivationKey } from "$utils";
 
 interface RippleOptions {
+  /**
+   * Whether the ripple should be centered on the element.
+   */
   center?: boolean;
-  color?: string; // CSS color
-  duration?: number; // In ms
-  disabled?: boolean; // Should the ripple be disabled
+  /**
+   * The color of the ripple. This should be in a format that can be used in CSS, such as hex, rgb, hsl, or named colors.
+   *
+   * For variables, no need to wrap in `var(--xx)` e.g. pass `primary` not `var(--primary)`.
+   */
+  color?: string;
+  /**
+   * The duration of the ripple in milliseconds.
+   */
+  duration?: number;
+  /**
+   * Whether the ripple should be disabled.
+   */
+  disabled?: boolean;
 }
 
-const triggerEvents = ["pointerdown", "touchstart", "keydown"] as const;
-const cancelEvents = ["mouseleave", "dragleave", "touchmove", "touchcancel", "pointerup", "keyup"];
+const TRIGGER_EVENTS = ["pointerdown", "touchstart", "keydown"] as const;
+const CANCEL_EVENTS = ["mouseleave", "dragleave", "touchmove", "touchcancel", "pointerup", "keyup"];
 
-export function ripple(el: HTMLElement, options: RippleOptions = {}) {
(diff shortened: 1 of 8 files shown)


src/lib/helpers/ripple.ts
```diff
@@ -1,70 +1,95 @@
...
+const TRIGGER_EVENTS = ["pointerdown", "touchstart", "keydown"] as const;
+const CANCEL_EVENTS = ["mouseleave", "dragleave", "touchmove", "touchcancel", "pointerup", "keyup"];
+
+export function ripple(options: RippleOptions = {}): Attachment<HTMLElement> {
+  const RIPPLE_CLASS = "q-ripple";
+  const CENTER_CLASS = `${RIPPLE_CLASS}--center`;
+  const EFFECT_CLASS = `${RIPPLE_CLASS}__effect`;
+
+  const COLOR_PROPERTY = "--ripple-color";
+  const DURATION_PROPERTY = "--ripple-duration";
```
therealcoder1337: You can put these outside of the function, since they don't depend on any input. And that might have a marginal positive effect on performance too, maybe.

src/lib/helpers/ripple.ts
```diff
@@ -1,70 +1,95 @@
...
+    if (opts.color) {
+      rippleContainer.style.setProperty(COLOR_PROPERTY, `var(--${opts.color}, ${opts.color})`);
     }
 
-    if (options.duration) {
-      rippleContainer.style.setProperty("--ripple-duration", `${options.duration}ms`);
+    if (opts.duration) {
+      rippleContainer.style.setProperty(DURATION_PROPERTY, `${opts.duration}ms`);
     }
   }
```
therealcoder1337, line 74: nit: We might want to move the duration part before the color part, as duration is already handled before color and thus it'll be a bit easier to read.

src/lib/helpers/ripple.ts
```diff
@@ -73,28 +98,26 @@ export function ripple(el: HTMLElement, options: RippleOptions = {}) {
...
-    const clientX =
-      window.TouchEvent && e instanceof TouchEvent
-        ? e.touches[0].clientX
-        : (e as PointerEvent).clientX;
-    const clientY =
-      window.TouchEvent && e instanceof TouchEvent
-        ? e.touches[0].clientY
-        : (e as PointerEvent).clientY;
+    const { clientX, clientY } =
+      window.TouchEvent && event instanceof TouchEvent ? event.touches[0] : (event as PointerEvent);
```
therealcoder1337, line 102: Does this work with SSR?

src/lib/helpers/ripple.ts
KilDesu: Done.

src/lib/helpers/ripple.ts
KilDesu: It should. As attachments run after mount (because they depend on DOM elements), I don't think this code runs on the server at all.

therealcoder1337 (approved):

therealcoder1337 (approved):

PR #211, opened 2026-06-14
feat: type widening

therealcoder1337:
## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

## What's new?

> List the changes

- feat(QInput): allow `null` value
- feat(QSelect): allow `null` value
- feat(QRadio): allow `number`, `boolean` and `null` value

These alignments make it possible to use these components a bit more flexible, without having too loose types like `any`.

## Screenshots

N/A

N/A

### DIFF
--- a/src/lib/components/radio/props.ts
+++ b/src/lib/components/radio/props.ts
@@ -1,12 +1,14 @@
 import type { HTMLAttributes } from "svelte/elements";
 
+export type QRadioValue = string | number | boolean | null;
+
 export interface QRadioProps extends HTMLAttributes<HTMLLabelElement> {
   /**
    * Value associated with this radio button. Used when comparing against the selected value.
    *
    * @default ""
    */
-  value: string;
+  value: QRadioValue;
 
   /**
    * Text label displayed next to the radio button.
@@ -21,7 +23,7 @@ export interface QRadioProps extends HTMLAttributes<HTMLLabelElement> {
    *
    * @default undefined
    */
-  selected?: unknown;
+  selected?: QRadioValue;
 
   /**
    * When true, prevents user interaction with the radio button.
--- a/src/lib/components/select/QSelect.svelte
+++ b/src/lib/components/select/QSelect.svelte
@@ -11,7 +11,7 @@
     getOptionLabel,
     normalizeOptionIndex,
   } from "./option";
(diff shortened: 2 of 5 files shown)


src/lib/components/radio/props.ts
```diff
@@ -1,12 +1,14 @@
 import type { HTMLAttributes } from "svelte/elements";
 
+export type QRadioValue = string | number | boolean | null;
```
KilDesu, line 3: Is `boolean` really indispensable here? Imho, I don't think boolean values make sense for radio buttons as they are basically useless on their own (especially when they can't be ticked off).
If you really want to keep it we can, but I personally don't think it's needed.

src/lib/components/select/QSelect.svelte
```diff
@@ -253,9 +253,11 @@
   }
 
   function isSelected(option: QSelectOption) {
+    const isArray = Array.isArray(value);
+
     return multiple
-      ? Array.isArray(value) && value.some((opt) => doValuesMatch(opt, option))
-      : doValuesMatch(value as QSelectOption, option);
+      ? isArray && (value as QSelectOption[]).some((opt) => doValuesMatch(opt, option))
+      : !isArray && doValuesMatch(value as QSelectOption, option);
```
KilDesu:
```suggestion
    const isArray = (val: QSelectValue): val is QSelectOption[] => Array.isArray(val);

   return multiple
      ? isArray(value) && value.some((opt) => doValuesMatch(opt, option))
      : !isArray(value) && doValuesMatch(value, option);
```
I haven't tried this code, but I feel like it's better than type casting. Take it or leave it.

KilDesu (approved):
Thanks! I added 2 comments but it's not blocking. You can address them or leave them as is.

src/lib/components/radio/props.ts
therealcoder1337: Yes, it's useful to have a boolean model sometimes, where you don't want a `QCheckbox`.

src/lib/components/select/QSelect.svelte
therealcoder1337: Thanks, applied!

PR #229, opened 2026-07-06
refactor: remove unused JSDoc, share props with centralized interface…

KilDesu:
…s and fix some problems

## PR Type

- [x] Bugfix
- [x] Feature
- [x] Documentation
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

## What's new?

> List the changes

- Fix some problems with props interface parsing
- Centralize similar props to be shared between multiple components
- Improve `getRouterInfo` to be more focused on links and use it where needed instead of repeating code
- Fix some UI problems
- Allow shared props to have a global default value for docs that can be overwritten by setting a default value inside the `.svelte` component file (e.g. `replace` prop is `false` in docs by default using `@default false` inside the interface. If a component defines `let { replace = true } = $props()`, the doc's default would become `true`)
- Improve some components' API using new shared props

## Screenshots

N/A

N/A

### DIFF
--- a/src/lib/utils/router.ts
+++ b/src/lib/utils/router.ts
@@ -33,7 +33,7 @@ export function getRouterInfo<T extends RouterProps>(props: T) {
       return linkAttributes;
     },
     get linkClass() {
-      return (hasLink && "q-link") || undefined;
+      return hasLink ? "q-link" : undefined;
     },
     get isActive() {
       return isActive;
--- a/src/lib/utils/router.ts
+++ b/src/lib/utils/router.ts
@@ -1,33 +1,42 @@
 import { page } from "$app/state";
+import type { Linkable, WithActiveAttrs } from "./types";
 
-export interface RouterProps {
-  activeClass?: string;
-  disabled?: boolean;
-  href?: string;
-  replace?: boolean;
-  to?: string;
-}
+interface RouterProps extends Linkable, WithActiveAttrs {}
 
 export function isRouteActive(route?: string) {
   if (!route) {
-    return false;
+    return;
   }
 
   return page.url.pathname === route || page.url.pathname.startsWith(`${route}/`);
 }
 
+/**
+ * Provides routing information for a component.
+ *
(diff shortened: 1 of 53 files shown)


src/lib/utils/types/base.ts
```diff
@@ -3,3 +3,9 @@ export type CssUnit = "px" | "%" | "em" | "ex" | "ch" | "rem" | "vw" | "vh" | "v
 export type CssValue = `${number}${CssUnit}`;
 
 export type StringWithSuggestions<T extends string> = T | (string & {});
+
+/**
+ * Utility type used to indicate the default value for a prop.
+ * This is used for decgen to know what the default value is for extended properties.
```
therealcoder1337:
```suggestion
 * This is used for docgen to know what the default value is for extended properties.
```

src/lib/utils/router.ts
```diff
@@ -1,33 +1,42 @@
...
-    linkAttributes,
-    linkClasses,
+    get hasLink() {
+      return hasLink;
+    },
+    get linkAttributes() {
+      return linkAttributes;
+    },
+    get linkClass() {
+      return (hasLink && "q-link") || undefined;
```
therealcoder1337:
```suggestion
      return hasLink ? "q-link" : undefined;
```
Let's use `&&` only when there's no better alternative.

src/lib/utils/types/base.ts
KilDesu: I actually forgot I defined this utility type. It's useless now because I found another solution so I removed it in [c05b0b1](https://github.com/quaffui/quaff/pull/229/commits/c05b0b19d4815104aa4260dd9a4de830a399eab4).

src/lib/utils/router.ts
KilDesu: Done in [c05b0b1](https://github.com/quaffui/quaff/pull/229/commits/c05b0b19d4815104aa4260dd9a4de830a399eab4).

therealcoder1337 (approved):

PR #233, opened 2026-07-11
feat: fixes and refactors

therealcoder1337:
## PR Type

- [x] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes
- [ ] No

## What's new?

- Fix `QSwitch` state styling and interactions, preserve its public `toggle()` method, and simplify its state-dependent CSS.
- Improve `QTable` with formatting, custom comparators, reliable sorting and pagination, accessible sort controls, and remove the unsupported `required` column option.
- Simplify `QCodeBlock`, correctly handle clipboard failures, and safely escape code when syntax highlighting fails.
- Generate color utility and semantic classes from centralized Sass collections instead of hundreds of repeated declarations.
- Remove unused legacy stylesheets for `QAvatar` and `QTabs`.

## Screenshots

N/A

N/A


src/lib/components/switch/QSwitch.scss
```diff
@@ -1,37 +1,12 @@
...
-      @return if(sass($hovered): var(--primary-container) ; else: var(--on-primary));
-    } @else {
-      @return if(sass($hovered): var(--on-surface-variant) ; else: var(--outline));
-    }
-  }
-
-  @error "The given element #{$el} is not supported.";
-}
+$checked: "[aria-checked=true]";
+$unchecked: "[aria-checked=false]";
```
KilDesu, line 9: What about if `aria-checked` is undefined? Would the selector work?

src/lib/components/codeBlock/QCodeBlock.svelte
```diff
@@ -82,34 +73,29 @@
 <div {...props} class="q-code-block" data-quaff>
   {#if copiable}
     <div
-      class="q-code-block__title-section justify-between {title
-        ? 'items-center'
-        : 'justify-end'} q-pb-sm"
+      class="q-code-block__title-section q-pb-sm"
+      class:items-center={title}
+      class:justify-between={title}
+      class:justify-end={!title}
```
KilDesu, line 79: Svelte advises to avoid using `class:` directive as it's deprecated. Maybe use an array or object?

src/lib/components/switch/QSwitch.scss
KilDesu, line 1: Are we sure we don't lose styles with the new code? I have no way to check atm.
Ignore if you're sure.

src/lib/components/switch/QSwitch.svelte
```diff
@@ -83,16 +84,11 @@
...
   data-quaff
 >
   <label class="q-switch__inner">
-    <input
-      bind:checked={value}
-      bind:this={qSwitchInput}
-      type="checkbox"
-      disabled={disabled || undefined}
-    />
+    <input bind:checked={value} bind:this={qSwitchInput} type="checkbox" {disabled} />
```
KilDesu, line 91: Are we sure we want `disabled=false` explicitly? I think it's better if we have it to be undefined but not too important

KilDesu (approved):
Some comments, but nothing blocking. Thanks!

src/lib/components/switch/QSwitch.scss
therealcoder1337: Shouldn't be `undefined`; `aria-checked={!!value}`

src/lib/components/switch/QSwitch.svelte
therealcoder1337: Hmm good point, will check.

therealcoder1337: > Boolean attributes are included on the element if their value is [truthy](https://developer.mozilla.org/en-US/docs/Glossary/Truthy) and excluded if it's [falsy](https://developer.mozilla.org/en-US/docs/Glossary/Falsy).

https://svelte.dev/docs/svelte/basic-markup

src/lib/components/codeBlock/QCodeBlock.svelte
therealcoder1337: It's not deprecated.

src/lib/components/switch/QSwitch.scss
therealcoder1337: QSwitch looks fine on that branch.

PR #234, opened 2026-07-11
perf: component runtime optimizations

therealcoder1337:
## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

## What's new?

> List the changes

- Optimize QSelect filtering and selected-option lookups for large option sets.
- Coalesce QMenu scroll-position updates into animation frames.
- Limit partial QTheme updates to the changed CSS variables.
- Lazily create ripple elements and skip setup for explicitly disabled ripples.
- Remove unnecessary reactive state from QDrawer swipe bookkeeping.
- Remove unnecessary reactive state from QTooltip controller bookkeeping.

## Screenshots

N/A

N/A

### DIFF
--- a/src/lib/components/tooltip/QTooltip.svelte
+++ b/src/lib/components/tooltip/QTooltip.svelte
@@ -28,17 +28,13 @@
   let removeTooltipMouseLeaveListener: (() => void) | undefined;
   let removeWindowWheelListener: (() => void) | undefined;
 
+  let tooltipEl: HTMLDivElement | undefined;
+  let realTarget: HTMLElement | undefined;
+  let timerShow: ReturnType<typeof setTimeout> | null = null;
+  let timerHide: ReturnType<typeof setTimeout> | null = null;
+
   const id = $props.id();
   // #endregion: --- Non-reactive variables
-
-  // #region:    --- Reactive variables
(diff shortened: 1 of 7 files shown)


src/lib/components/tooltip/QTooltip.svelte
```diff
@@ -28,18 +28,14 @@
   let removeTooltipMouseLeaveListener: (() => void) | undefined;
   let removeWindowWheelListener: (() => void) | undefined;
 
+  let tooltipEl: HTMLDivElement | undefined;
```
KilDesu, line 31: I think we need the `$state` here, Svelte says to use it with `bind:this`.
Can you check this?

KilDesu (approved):
Can you please check my comment before merging? Thanks.

src/lib/components/tooltip/QTooltip.svelte
therealcoder1337: I didn't see any warning about this, and the local docs page works fine.
There's however a bound var by the same name in `QTooltipBase`, maybe you're confusing these?

PR #235, opened 2026-07-12
refactor: q-header to better match md3 app-bar

### LINKED ISSUE #214: QHeader refactor & QToolbar removal
- QToolbar should be removed, as it could be confused with the MD3 Toolbar component we have not yet implemented
- QHeader should take over the role & functionalities of the current QToolbar
- Also replace or rename the then-obsolete QToolbarTitle

This is **not** about implementing the MD3 Toolbar. That might be done as part of a future task, if we choose to have it.

therealcoder1337:
## PR Type

- [x] Bugfix
- [ ] Feature
- [x] Documentation
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes
- [ ] No

`QToolbar` and `QToolbarTitle` are removed in favor of using `QHeader` with the new `QHeaderTitle`. Buttons also no longer provide implicit horizontal margins.

## What's new?

> List the changes

- Refactor `QHeader` to provide the app-bar layout directly, removing the redundant toolbar components
- Remove implicit horizontal margins from `QBtn`
- Fix rounded `QInput` and `QSelect` borders incorrectly showing a floating-label notch when no label is present

## Screenshots

N/A

Closes #214.


src/lib/components/header/QHeader.scss
```diff
@@ -3,23 +3,36 @@
...
   border-radius: 0;
 
   background-color: var(--surface);
+  transition: all variables.$speed3;
+
+  > .q-btn--round {
+    width: 3rem;
+    min-width: 3rem;
+    height: 3rem;
+  }
```
KilDesu, line 25: With the expressive PR, the round prop will not exist anymore. I don't think this will ever select anything in the future.

KilDesu (approved):
Thanks!

therealcoder1337:
The e2e failure is expected for now, we'll run e2e after create-quaff has been updated, ensuring it's green before publishing new releases.

PR #252, opened 2026-07-19
feat(QBtn,QIconBtn): align with md3 specs

therealcoder1337:
## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

## What's new?

> List the changes

- QBtn,QIconBtn: align with md3 specs

## Screenshots

N/A

N/A

### DIFF
--- a/src/lib/components/button/QBtn.scss
+++ b/src/lib/components/button/QBtn.scss
@@ -32,6 +32,17 @@
       variables.$speed3 background-color,
       variables.$speed3 border-radius;
 
+    // Target area so small buttons remain accessible
+    &::before {
+      content: "";
+      position: absolute;
+      top: 50%;
+      left: 50%;
+      width: max(100%, 3rem);
+      height: max(100%, 3rem);
+      transform: translate(-50%, -50%);
+    }
+
     &:not(:has(> .q-btn__label)) {
       width: var(--q-btn-height, 2.5rem);
       padding-inline: 0;
@@ -94,6 +105,7 @@
 
       &.q-btn--lg {
         --q-btn-height: 6rem;
+        --q-btn-outline-width: 0.125rem;
         padding-inline: 3rem;
         gap: 0.75rem;
         @include mixins.typography(headline-small);
@@ -101,12 +113,13 @@
 
       &.q-btn--xl {
         --q-btn-height: 8.5rem;
+        --q-btn-outline-width: 0.1875rem;
         padding-inline: 4rem;
         gap: 1rem;
         @include mixins.typography(headline-large);
       }
 
-      &.q-btn--squared {
+      &:is(.q-btn--squared, .q-btn--selected) {
         &.q-btn--xs,
         &.q-btn--sm {
           border-radius: 0.75rem;
@@ -120,6 +133,10 @@
         &.q-btn--xl {
           border-radius: 1.75rem;
         }
+      }
+
+      &.q-btn--selected.q-btn--squared {
+        border-radius: calc(var(--q-btn-height) / 2);
       }
 
       &:active {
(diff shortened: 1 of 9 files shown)


src/lib/components/button/QBtn.scss
```diff
@@ -32,6 +32,16 @@
       variables.$speed3 background-color,
       variables.$speed3 border-radius;
```
KilDesu, line 34:
```suggestion

// Target area so small buttons remain accessible
```
Can we add this comment if this code is indeed for the target area (see https://m3.material.io/components/buttons/specs#7d63205e-9401-449b-9a2b-3408f4de3b0d)?

src/lib/components/button/QBtn.scss
```diff
@@ -159,23 +175,46 @@
...
-        @include mixins.elevate(1, "bottom");
-      }
-    }
-
     &--outlined {
       background-color: transparent;
       --ripple-color: var(--on-surface-variant);
-
-      @include mixins.border();
+      border: var(--q-btn-outline-width, 0.0625rem) solid var(--outline-variant);
```
KilDesu:
```suggestion
      @include mixins.border($width: var(--q-btn-outline-width, 0.0625rem));
```
Can we change the `border` mixin to accept a `$width` parameter that defaults to `0.0625rem` so we can have the above code?

src/lib/components/button/QBtn.scss
KilDesu, line 1: When a selected button is in the pressed (active) state, it should have the same border-radius as a non-selected button in the same state. Right now, the border radius change on press works for non-selected buttons (image 1), but not for selected ones (image 2). Both images show buttons in their pressed (active) state.

Image

Image

src/lib/components/button/QBtn.scss
```diff
@@ -212,12 +251,13 @@
   .q-btn[aria-disabled] {
     @include disabled.base();
 
-    &:not(.q-btn--outlined, .q-btn--flat) {
-      background-color: disabled.rest-color();
+    &:not(.q-btn--outlined, .q-btn--flat),
+    &.q-btn--outlined.q-btn--selected {
+      background-color: mixins.with-alpha(var(--on-surface), 10%);
```
KilDesu:
```suggestion
      background-color: disabled.rest-color($opacity: 10%);
```
Can we change the `rest-color` mixin to accept a `$opacity` parameter that defaults to `12%` so we can have the above code?
This will also serve for inputs which have a background opacity of `4%` when disabled.

src/lib/components/button/QBtn.svelte
KilDesu, line 1: MD3 seems to use filled icons for selected buttons, can we add `filled={selected}` to `QIconSnippet`?
Actually, they seem to use filled icons for both standard and toggle+selected buttons, maybe we can add `filled={selected === undefined || selected}`? The first suggestion is important but the second one not so much, we can discuss it or drop it.

Image

src/routes/components/button/+page.svelte
```diff
@@ -128,6 +132,20 @@
...
+          <code>aria-pressed</code> automatically. The text variant is not available for toggles.
+        </p>
+        <QBtn class="q-ma-sm" expressive bind:selected={filledSelected} label="Filled" filled />
+        <QBtn
+          class="q-ma-sm"
+          expressive
+          bind:selected={outlinedSelected}
+          label="Outlined"
+          outlined
+        />
```
KilDesu: This should be in its own `<QDocsSection>`, especially since our toggle implementation is apparently not limited to `expressive` buttons (which MD3 seem to do).

KilDesu (changes requested):
Thanks!

src/lib/components/button/QBtn.svelte
KilDesu: Actually it seems that text (flat) buttons can't be toggle buttons. But flat icon buttons can:



What should we do about that?

src/lib/components/button/QBtn.scss
therealcoder1337: Done!

src/lib/components/button/QBtn.scss
therealcoder1337: Done.

src/lib/components/button/QBtn.scss
therealcoder1337: Done.

src/lib/components/button/QBtn.svelte
therealcoder1337: Implemented with the spec distinction: default labeled and selected-toggle icons are filled, while default icon-only and unselected-toggle icons remain outlined.

src/routes/components/button/+page.svelte
therealcoder1337: Done.

src/lib/components/button/QBtn.scss
therealcoder1337: Please check again.

KilDesu (approved):

PR #266, opened 2026-07-25
fix: fix various bugs

therealcoder1337:
## PR Type

- [x] Bugfix
- [ ] Feature
- [x] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

## What's new?

> List the changes

A bunch of AI assorted and validated (GPT 5.6 Sol Ultra) fixes.

- **router: apply replace-state navigation**  
  The `replace` prop added the reload attribute instead of the replace-state attribute. The old version reloaded the page, while the fixed version replaces the current history entry.

- **preprocessor: interpolate class namespace errors**  
  Preprocessor errors displayed the literal text `${namespace}`. They now show the actual namespace that caused the error.

- **QScrollObserver: initialize observed scroll state**  
  An observer mounted at scroll position 200 incorrectly started at zero. Its first upward movement was reported as downward; it now starts from the real position and reports the correct direction and delta.

- **QBtn: respect prevented interactions**  
  Calling `preventDefault()` in a button handler did not stop toggle behavior. Consumer handlers now run first, so canceled interactions no longer change the button state.

- **QChip: preserve consumer interaction handlers**  
  Chips changed state before consumer click and keyboard handlers could cancel the interaction. The handlers now run first and prevented interactions leave the chip unchanged.

- **QDialog: preserve consumer dismissal handlers**  
  QDialog replaced consumer click, keyboard, and cancel handlers with its internal handlers. These handlers are now called, and preventing Escape keeps the dialog open while normal outside clicks still close it.

- **QDrawer: keep persistent modal drawers open**  
  Persistent drawers could still close through the modal scrim. The scrim can no longer dismiss a persistent drawer and is removed from keyboard interaction.

- **QExpansionItem: preserve link and toggle behavior**  
  Link options such as `target` were not forwarded, and internal navigation overrode native link behavior. Link attributes now work normally, and prevented keyboard events no longer toggle the item.

- **QFooter: remove hidden footer layout space**  
  A hidden footer could still reserve its height in the layout. Hiding it now also removes that unused space.

- **QIcon: support image icon names**  
  Supported `img:` icon names were rendered as plain text. They now produce an image element using the extracted source.

- **QMenu: preserve consumer dismissal handling**  
  Menu click handlers could not prevent automatic closing, and Escape was handled before child elements could cancel it. Prevented click and Escape events now keep the menu open.

- **QSelect: forward field interaction attributes**  
  Field attributes such as `aria-label` were not applied to the actual select input, and keyboard behavior ran before consumer handlers. Attributes now reach the input and prevented keyboard events stop internal handling.

- **QSwitch: respect prevented interactions**  
  QSwitch toggled even when its click or keyboard handler canceled the event. Prevented interactions now leave its value unchanged.

- **QTable: label pagination controls**  
  Pagination controls had no accessible names. The rows-per-page select and navigation buttons now have clear labels.

- **QTable: support numeric field resolvers**  
  QTable rows allow numbers, but field resolver functions were typed to return only strings. Numeric resolver results are now accepted consistently.

- **QTab: respect prevented keyboard events**  
  Preventing a keyboard event did not stop tab activation. QTab now checks the canceled state before activating the tab.

- **QTabs: discover dynamic tab elements**  
  The cached tab list did not update when tabs were added after mounting. QTabs now reads the current tab elements when it needs them, allowing dynamic tabs to work.

- **useAlign: apply vertical alignments**  
  Values such as `top`, `middle`, and `bottom` were accepted but produced no vertical alignment classes. They now map to the corresponding cross-axis alignment classes.

- **QColors: support short hex conversion**  
  QColors accepted short hex values such as `#abc`, but converted them to empty or invalid results. Short values are now expanded and converted correctly.

- **agents: require focused and readable changes**  
  The contributor guidance now requires every diff to have a clear reason. It also documents the expected blank-line style around control-flow blocks.

## Screenshots

N/A

N/A

### DIFF
--- a/src/lib/classes/QScrollObserver.svelte.ts
+++ b/src/lib/classes/QScrollObserver.svelte.ts
@@ -69,14 +69,20 @@ export default class QScrollObserver {
     this.direction = horizontal ? "right" : "down";
 
     onMount(() => {
-      let parsedTarget = elFromSelector(target ?? null);
+      const resolvedTarget =
+        target === null || target === undefined ? window : elFromSelector(target);
 
-      if (parsedTarget === null) {
+      if (resolvedTarget === null) {
         console.warn(`The given target (${target}) is null, observing window instead`);
-        parsedTarget = window;
       }
 
+      const parsedTarget = resolvedTarget instanceof HTMLElement ? resolvedTarget : window;
+
       this.target = parsedTarget;
+      const initialPosition = this.getScrollPosition(parsedTarget);
+      this.lastScroll = initialPosition;
+      this.position = initialPosition;
+      this.inflectionPoint = initialPosition;
 
       this.removeScrollListener = on(parsedTarget, "scroll", this.handler, SCROLL_LISTENER_OPTIONS);
 
@@ -131,6 +137,11 @@ export default class QScrollObserver {
     this.position = newScrollPosition;
     this.delta = newScrollPosition - this.lastScroll;
 
+    if (this.delta === 0) {
+      this.changedDirection = false;
+      return;
+    }
+
     const newDirection: Direction =
       this.delta > 0 ? (this.horizontal ? "right" : "down") : this.horizontal ? "left" : "up";
 
@@ -165,7 +176,7 @@ export default class QScrollObserver {
     this.removeScrollListener = null;
 
     this.target = null;
-    this.direction = "down";
+    this.direction = this.horizontal ? "right" : "down";
     this.changedDirection = false;
     this.delta = 0;
     this.inflectionPoint = 0;
--- a/src/lib/components/footer/QFooter.svelte
+++ b/src/lib/components/footer/QFooter.svelte
@@ -28,14 +28,14 @@
   // #endregion: --- Reactive variables
 
   // #region:    --- Derived values
-  const revealObserver = useRevealScrollObserver("footer", uid, () => reveal);
+  const revealObserver = useRevealScrollObserver("footer", uid, () => reveal && value);
   const revealScroll = $derived(revealObserver.scroll);
 
   const offset = $derived(revealScroll ? revealScroll.position + height : undefined);
 
   // Collapse the footer `${revealOffset}px` above the bottom of layout content when scrolling up
   const collapsed = $derived(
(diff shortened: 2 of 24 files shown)


src/lib/classes/QScrollObserver.svelte.ts
```diff
@@ -69,14 +69,19 @@ export default class QScrollObserver {
     this.direction = horizontal ? "right" : "down";
 
     onMount(() => {
-      let parsedTarget = elFromSelector(target ?? null);
+      const resolvedTarget = target == null ? window : elFromSelector(target);
```
KilDesu: Why only `==` and not `===`?

src/lib/components/footer/QFooter.svelte
```diff
@@ -28,14 +28,15 @@
...
   const revealScroll = $derived(revealObserver.scroll);
 
   const offset = $derived(revealScroll ? revealScroll.position + height : undefined);
 
   // Collapse the footer `${revealOffset}px` above the bottom of layout content when scrolling up
   const collapsed = $derived(
-    reveal && revealScroll?.direction === "up" && offset! + revealOffset < contentScrollHeight
+    !value ||
+      (reveal && revealScroll?.direction === "up" && offset! + revealOffset < contentScrollHeight)
   );
```
KilDesu, line 40: Can we simplify this?

src/lib/components/icon/QIcon.svelte
KilDesu, line 1: Not sure we should have 2 ways to have an image icon (2 different props). What about deleting the `` | `img:${string}` `` from the props file?

KilDesu (approved):
Approved but some comments need to be addressed first pls

src/lib/components/chip/props.ts
KilDesu, line 1: Seems that `QIcon` needs to support `img:` so `QChip` can too.
Could you reimplement it so `QChip`'s API doesn't suffer from it?

KilDesu (changes requested):

KilDesu (approved):

PR #291, opened 2026-09-13
feat(QBreadcrumbs): align with md3 specs

therealcoder1337:
## PR Type

- [x] Bugfix
- [x] Feature
- [x] Documentation
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

## What's new?

- feat(QBreadcrumbs): align with md3 specs

## Screenshots

N/A

N/A


src/lib/components/breadcrumbs/QBreadcrumbs.svelte
```diff
@@ -43,24 +33,18 @@
   });
   // #endregion: --- Context
 
-  // #region:    --- Lifecycle
-  onMount(() => {
-    breadcrumbList.querySelector(".q-breadcrumbs__separator:first-child")?.remove();
-  });
-  // #endregion: --- Lifecycle
-
```
KilDesu, line 51: This will make it so the first breadcrumb inside *all* `QBreadcrumbs` will have a hidden separator that's not removed, meaning extra DOM elements for reason.
I can refactor similar components in a later PR so we don't have to remove elements in `onMount` for them (e.g. `QItem` inside separated lists).

KilDesu (approved):

src/lib/components/breadcrumbs/QBreadcrumbs.svelte
therealcoder1337: True, but let's keep it simple for now.

PR #293, opened 2026-09-13
feat(QLayout): align with md3 specs & QDrawer fix

### LINKED ISSUE #286: Fix QLayout
There's an issue with `QLayout`'s right drawer in the docs example, this needs to be investigated and fixed.

therealcoder1337:
## PR Type

- [x] Bugfix
- [x] Feature
- [x] Documentation
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

## What's new?

- QLayout: align with md3 specs
- QDrawer: keep bordered drawers square

Follows [Material’s adaptive layout guidance](https://developer.android.com/codelabs/adaptive-material-guidance) by fitting the available space and updating content offsets as navigation changes. Layout transitions respect reduced motion following [WCAG guidance](https://www.w3.org/WAI/WCAG22/Understanding/animation-from-interactions.html). Bordered standard drawers use square corners so dividers meet adjacent bars cleanly, consistent with the [Material 3 permanent drawer default](https://developer.android.com/reference/kotlin/androidx/compose/material3/PermanentDrawerSheet.composable).

## Screenshots

N/A

Closes #286.


src/lib/components/drawer/QDrawer.svelte
```diff
@@ -46,8 +46,9 @@
   // #endregion: --- Reactive variables
 
   // #region:    --- Derived values
-  const drawerCtxToUse = $derived(side === "left" ? leftDrawerCtx : rightDrawerCtx);
-  const drawerContext = $derived(drawerCtxToUse.get());
```
KilDesu, line 50: Feels like it would require more memory than the old implementation as we get both contexts and define 3 variables.

src/lib/components/drawer/QDrawer.svelte
```diff
@@ -149,16 +144,17 @@
...
+
+    if (!context) {
+      return;
+    }
+
+    Object.assign(context, { takesSpace: !!value && !isModal, width, ready: true });
+
+    return () => {
+      Object.assign(context, { takesSpace: false, width: 0, ready: false });
+    };
```
KilDesu, line 158: - Why defining `const context` just for the sake of an extra variable definition? It gives nothing more.
- Not sure about using `Object.assign`, I don't know if that works well with reactivity.

src/lib/components/layout/QLayout.svelte
```diff
@@ -162,56 +155,34 @@
...
-      !navbar || navbarInfo.ready || (layoutEl && !layoutEl.querySelector(":scope > .q-navbar"))
+      !snippet || info.ready || (!!layoutEl && !layoutEl.querySelector(`:scope > .${selector}`))
     );
   }
   // #endregion: --- Functions
 
+  onMount(() => {
+    let frame = requestAnimationFrame(() => {
+      frame = requestAnimationFrame(() => (isAnimated = true));
+    });
```
KilDesu, line 178: What does the nested animation frame do? Seems to replace the timeout but how does it work?
The timeout was a dirty hack to avoid having the whole layout awkwardly move with transitions on page load. Would that work properly? Would `tick` allow to do the same (while being a more svelte idiomatic way to do so)?

src/lib/components/railbar/QRailbar.svelte
```diff
@@ -15,59 +14,56 @@
   }: QRailbarProps = $props();
   // #endregion: --- Props
 
-  // #region:    --- Non-reactive variables
-  let railbarEl: HTMLElement;
-  // #endregion: --- Non-reactive variables
+  let railbarEl = $state<HTMLElement>();
+  const leftContext = leftRailbarCtx.get();
+  const rightContext = rightRailbarCtx.get();
```
KilDesu, line 19: Same comment as for `QDrawer`.

src/lib/components/railbar/QRailbar.svelte
```diff
@@ -15,59 +14,56 @@
...
+
+      if (!measuredWidth && style.display !== "none") {
+        measuredWidth = style.width.endsWith("px") ? parseFloat(style.width) : configuredWidth;
+      }
+
+      Object.assign(context, { width: measuredWidth, takesSpace: measuredWidth > 0, ready: true });
+    };
+    const observer = new ResizeObserver(updateLayout);
+    updateLayout();
+    observer.observe(element);
```
KilDesu, line 54: Same comment as for `QDrawer`.

KilDesu (commented):
Some comments

src/lib/components/drawer/QDrawer.svelte
therealcoder1337: These hold references to contexts QLayout already creates; they don’t duplicate the objects. Capturing both during initialization keeps later `side` changes from calling `getContext()` outside its supported lifecycle. The selected context remains derived.

src/lib/components/drawer/QDrawer.svelte
therealcoder1337: The local `context` captures the original side so cleanup resets that side when `side` changes. Reading `drawerContext` again could target the new side instead. Added a comment explaining this.

`Object.assign` invokes the reactive setters generated by our context preprocessor. `QContext.updateEntries()` already uses the same operation internally.

src/lib/components/layout/QLayout.svelte
therealcoder1337: The second animation frame gives the initial bar measurements and offsets a rendering opportunity before transitions become active. This addresses the initial animated shift without the arbitrary timeout.

`tick()` waits for Svelte’s DOM updates, but doesn’t wait for the browser’s rendering and ResizeObserver measurements. Added a comment explaining the timing; the pending frame is also cancelled on unmount.

src/lib/components/railbar/QRailbar.svelte
therealcoder1337: Same reasoning here: these are references to existing contexts, with both lookups performed during component initialization. The derived selection then handles `side` changes without another `getContext()` call.

src/lib/components/railbar/QRailbar.svelte
therealcoder1337: The captured context keeps the observer and cleanup tied to the side they were created for. Otherwise, changing `side` could leave the old side’s width behind. Added a comment explaining this.

`Object.assign` remains reactive through our generated setters, just as it was inside `QContext.updateEntries()`.

src/lib/components/drawer/QDrawer.svelte
KilDesu: > `QContext.updateEntries()` already uses the same operation internally.

So why not use `QContext.updateEntries`?

src/lib/components/railbar/QRailbar.svelte
KilDesu: Idem, why not use our own internal API?

KilDesu (approved):

PR #297, opened 2026-09-15
feat(QFooter): align with md3 specs

therealcoder1337:
## PR Type

- [x] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

## What's new?

- feat(QFooter): align with md3 specs

Uses Quaff’s existing `body-medium` typography from the [MD3 type scale](https://m3.material.io/styles/typography/type-scale-tokens), with a `surface` background and inherited text color appropriate for a generic layout footer. [WCAG focus guidance](https://www.w3.org/WAI/WCAG22/Understanding/focus-not-obscured-minimum.html) informs revealing the footer when its content receives keyboard focus, and [animation guidance](https://www.w3.org/WAI/WCAG22/Understanding/animation-from-interactions.html) informs respecting reduced-motion preferences.
Also fixes the `reveal` animation.

## Screenshots

N/A

N/A


src/lib/components/footer/QFooter.scss
```diff
@@ -11,7 +11,10 @@
 
   @include mixins.layoutBarDisplay;
 
-  @include mixins.padding("x-md");
```
KilDesu, line 14: This is equivalent to `padding-inline: 1rem;` so we can keep it instead of it.

KilDesu (approved):
One comment, but not blocking

src/lib/components/footer/QFooter.scss
therealcoder1337: Let's do it on a later CSS refactor.

PR #299, opened 2026-09-15
fix(QList): allow native tab order

therealcoder1337:
## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

## What's new?

- QList: allow native tab order

[Material 3 list accessibility](https://m3.material.io/components/lists/accessibility) defines arrow-key navigation within lists, so this remains the default. `QApi` uses `preserveTabOrder` to allow normal Tab navigation through clickable types. `QSelect` also uses it to retain its existing input focus management, following the [WAI-ARIA combobox pattern](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/).

## Screenshots

N/A

N/A


src/lib/components/list/props.ts
```diff
@@ -37,6 +37,12 @@ export interface QListProps extends WithActiveAttrs, HTMLAttributes<HTMLElement>
    */
   selection?: QListSelection;
 
+  /**
+   * Preserves child elements' tab order and disables the list's arrow-key navigation.
+   * Defaults to false.
```
KilDesu: Do we really need to explicit the default value?

KilDesu (approved):

src/lib/components/list/props.ts
therealcoder1337: Fixed.

PR #300, opened 2026-09-16
feat(QTable): align with md3 specs

therealcoder1337:
## PR Type

- [x] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

## What's new?

- feat(QTable): align with md3 specs

[Google’s MD3 table tokens](https://github.com/material-components/material-web/blob/a300d043d2e11886b0be3c0ee429c6bd336c41e9/tokens/versions/latest/sass/_md-comp-data-table.scss#L23-L110) define the dimensions, colors and shapes, using Quaff’s existing `title-small` and `body-medium` typography. Dense sizing follows [Angular Material’s density mapping](https://github.com/angular/components/blob/7e73582615a66007462bd91496720d822b720b56/src/material/table/_m3-table.scss#L42-L51), while the footer retains enough space for its controls. [WAI’s sortable-table guidance](https://www.w3.org/WAI/ARIA/apg/patterns/table/examples/sortable-table/) informed the native header buttons, `aria-sort`, and visible keyboard focus.

## Screenshots

N/A

N/A


src/lib/components/table/QTable.svelte
```diff
@@ -13,47 +13,52 @@
...
+  let sortName = $state<string>();
+  let isDescending = $state(false);
   // #endregion: --- Reactive variables
 
   // #region:    --- Derived values
   const rowsPerPageOptions = $derived(
-    [5, 10, 25, 50].filter((option) => rows.length >= option || option === 5)
+    [5, 10, 25, 50].filter(
+      (option) => rows.length >= option || option === rowsPerPage || option === 5
+    )
```
KilDesu: Tbh, I find the `rows.length >= option` a bit weird.
Imagine you have 23 rows: you're stuck with only 10 rows visible so you have 3 pages to navigate (seems like not very good UX).
Honestly, I think all the options should be visible. Wdyt?

src/lib/components/table/QTable.svelte
```diff
@@ -76,138 +81,139 @@
...
+              {@const columnCell = props[`bodyCell${capitalize(column.name)}`]}
+              {@const cell = typeof columnCell === "function" ? columnCell : bodyCell}
+              {#if cell}
+                {@render cell({ column, row, style: getCellStyle(column) })}
+              {:else}
+                <td class="q-table__body-cell" style={getCellStyle(column)}>
+                  {getCellValue(column, row)}
+                </td>
+              {/if}
+            {/each}
```
KilDesu, line 167:
```suggestion
            {#each columns as column (column.name)}
              {@const style = getCellStyle(column)}
              {@const columnCell = props[`bodyCell${capitalize(column.name)}`]}
              {@const cell = typeof columnCell === "function" ? columnCell : bodyCell}
              {#if cell}
                {@render cell({ column, row, style })}
              {:else}
                <td class="q-table__body-cell" {style}>
                  {getCellValue(column, row)}
                </td>
              {/if}
            {/each}
```

KilDesu (changes requested):
Thanks! Some comments that need to be addressed

KilDesu (approved):

PR #307, opened 2026-09-17
refactor(QContext): use captured context references

### LINKED ISSUE #296: Refactor `QContext`
We should probably refactor `QContext` to leverage Svelte reactivity and context API better.

Today, `QContext.updateEntry` (and `updateEntries`) call `getContext` on every call, which is not desirable as those are usually called inside effects.

therealcoder1337:
## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes
- [ ] No

Removes the undocumented `updateEntry` and `updateEntries` helpers. Code using them must retrieve the context during component initialization and update it directly.

## What's new?

- QContext: use captured context references

Headers and footers now reuse their existing context, fixing context lookup errors when components are removed. Replacing a bar or interrupting its exit transition also preserves the correct spacing, preventing the outgoing bar from clearing its replacement’s layout state.

## Screenshots

N/A

Closes #296


src/lib/utils/context.ts
KilDesu, line 1: That's really not the changes I expected, I do not want to delete everything. Now, it is pretty much useless.

I was thinking more of allow passing the context to `updateEntry` and `updateEntries` instead of calling `getContext` inside them.
This way, components can just call `const ctx = QContext.get();` to load the context once, and then can modify it with `QContext.updateEntry(ctx, "myKey", "myValue")`.
I just want to sugarcoat those ugly `Object.assign` in the code.

KilDesu (changes requested):

KilDesu (approved):

